PR #2, opened 2016-07-26
(MODULES-3048) Add PowerShell Manager

jpogran:
Port PowerShell Manager from the PowerShell Module. Add subfolders and rename files to avoid file
name collisions

PR #1 needs to be merged before this PR


glennsarti:
:+1:

PR #3, opened 2016-07-26
(MODULES-3049) Add type and provider for IIS sites

jpogran:
Add a type and provider for listing, creating and deleting IIS Websites. This does not
handle setting binding information for a site, this is covered in MODULES-3639
- [x]  PowerShell WebAdministration ModuleProvider
- [x]  Create, Modify, Delete iis sites
- [x]  Service Auto Start Provider
- [x]  Unit tests for type and provider
- [x]  IIS version Fact
- [x]  Confines for IIS and OS version for 2016

### DIFF
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -1,0 +1,175 @@
+require 'pathname'
+
+# When writing IIS PowerShell code for any of the methods below
+# NEVER EVER use Get-Website without specifying -Name. As the number
+# of sites on a server increases, Get-Website will take longer and longer
+# to return. This will exponentially increase the total duration of your
+# puppet run
+
+Puppet::Type.type(:iis_site).provide(:webadministration) do
+  desc "IIS Provider using the PowerShell WebAdministration module"
+
+  require Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/puppetlabs/iis/powershell_manager'
+  require Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/puppetlabs/iis/powershell_common'
+  include PuppetX::IIS::PowerShellCommon
+  
+  confine    :iis_version     => ['8','8.5']
+  confine    :operatingsystem => [:windows ]
+  defaultfor :operatingsystem => :windows
+
+  commands :powershell => PuppetX::IIS::PowerShellCommon.powershell_path
+
+  mk_resource_methods
+
+  def create
+    cmd = []
+
+    cmd << self.class.ps_script_content('_newwebsite', @resource)
+
+    cmd << self.class.ps_script_content('generalproperties', @resource)
+
+    cmd << self.class.ps_script_content('logproperties', @resource)
+
+    cmd << self.class.ps_script_content('serviceautostartprovider', @resource)
+
+    inst_cmd = cmd.join
+
+    result = self.class.run(inst_cmd)
+
+    Puppet.err "Error creating website: #{result[:errormessage]}" unless result[:exitcode] == 0
+    Puppet.err "Error creating website: #{result[:errormessage]}" unless result[:errormessage].nil?
+
+    return exists?
+  end
+
+  def destroy
+    inst_cmd = "Remove-Website -Name \"#{@resource[:name]}\" -ErrorAction Stop"
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error destroying website: #{result[:errormessage]}" unless result[:exitcode] == 0
+    Puppet.err "Error destroying website: #{result[:errormessage]}" unless result[:errormessage].nil?
+    return exists?
+  end
+
+  def exists?
+    inst_cmd = "Get-Website -Name '\"#{@resource[:name]}\"'"
+    result   = self.class.run(inst_cmd)
+    
+    resp = result[:stdout]
+    if resp.nil?
+      return false
+    else
+      return true
+    end
+  end
+
+  def start
+    create if ! exists?
+    @resource[:ensure]  = 'started'
+
+    inst_cmd = "Start-Website -Name \"#{@resource[:name]}\""
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error starting website: #{result[:errormessage]}" unless result[:errormessage].nil?
+    
+    resp     = result[:stdout]
+    if resp.nil?
+      return true
+    else
+      return false
+    end
+  end
+
+  def stop
+    create if ! exists?
+    @resource[:ensure] = 'stopped'
+
+    inst_cmd = "Stop-Website -Name \"#{@resource[:name]}\""
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error stopping website: #{result[:errormessage]}" unless result[:errormessage].nil?
+    
+    resp = result[:stdout]
+    if resp.nil?
+      return true
+    else
+      return false
+    end
+  end
+
+  def initialize(value={})
+    super(value)
+    @property_flush = {}
+  end
+
+  def self.prefetch(resources)
+    sites = instances
+    resources.keys.each do |site|
+      if provider = sites.find{ |s| s.name == site }
+        resources[site].provider = provider
+      end
+    end
+  end
+
+  def self.instances
+    inst_cmd = ps_script_content('_getwebsites', @resource)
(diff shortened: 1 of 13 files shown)


glennsarti:
`lib/puppet/provider/iis_site/iisadministration.rb` is empty.  Is this a placeholder?

jpogran:
@glennsarti Yes, that is a placeholder. The functionality to support Windows Server 2016/Nano goes there

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -0,0 +1,182 @@
...
+
+    cmd << self.class.ps_script_content('serviceautostartprovider', @resource)
+
+    inst_cmd = cmd.join
+
+    result = self.class.run(inst_cmd)
+
+    if result[:exitcode] != 0
+      fail("Error creating website: #{result[:errormessage]}")
+      return false
```
MosesMendoza: I don't think we'll ever get to this `return` as `fail` will raise an exception and we'll return context to the caller.

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -0,0 +1,182 @@
...
+    end
+    Puppet.err "Error creating website: #{result[:errormessage]}" unless result[:errormessage].nil?
```
MosesMendoza, line 40: I think this (and the subsequent error loggings) may be more intuitive without the double-negative conditional, e.g., as `Puppet.err "Error creating website: #{result[:errormessage]}" if result[:errormessage]`

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -0,0 +1,182 @@
...
+
+    return exists?
+  end
+
+  def destroy
+    inst_cmd = "Remove-Website -Name \"#{@resource[:name]}\" -ErrorAction Stop"
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error destroying website: #{result[:errormessage]}" unless result[:errormessage].nil?
+    exists? ? (return true) : (return false)
```
MosesMendoza: could this line just be `exists?` ?

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  def exists?
+    inst_cmd = "Get-Website -Name \"#{@resource[:name]}\""
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error finding website: #{result[:errormessage]}" unless result[:errormessage].nil?
+    resp     = result[:stdout]
```
MosesMendoza: not sure about the conciseness/readability tradeoff but could this and the subsequent 5 lines be collapsed into `!!(result[:message])` ?

lib/puppet/provider/iis_site/webadministration.rb
glennsarti, line 40: I'm not a fan of doing implicit boolean conversion on string types.  e.g. what if the returned values is `false`.

lib/puppet/type/iis_site.rb
```diff
@@ -0,0 +1,270 @@
+require 'puppet/parameter/boolean'
+
+Puppet::Type.newtype(:iis_site) do
+  @doc = "Create a new IIS website."
+
+  newproperty(:ensure) do
+    desc "Specifies whether a site should be present or absent. If present is
+      specified, the site will be created but left in the default stopped state.
+      If started is specified, then the site will be created as well as started.
+      If stopped is specified, then the site will be created and kept stopped."
```
MosesMendoza, line 10: This is interesting - so a user could specify `ensure => present` in their manifest and the site could actually be in either the `started` or `stopped` state? (ie if the user manually started the site after puppet ran with `ensure => present`

glennsarti, line 10: Are you suggesting ensure should be `present|absent` and another property of, say, `status` be `started|stopped` ?

MosesMendoza, line 10: I'm not sure? another option would be to forgo `present|absent` and have only `started|stopped` - I believe this is what the `service` type does. e.g., if `ensure => stopped` and site doesn't exist, the equivalent of `present` is assumed.

glennsarti, line 10: Probably wouldn't work in this case.  `present and stopped` is not the same as `doesn't exist`

lib/puppet/provider/iis_site/webadministration.rb
MosesMendoza, line 40: Fair enough:) without any assurance that result[:errormessage] will be a string (thanks ruby) this could indeed behave differently than expected.

lib/puppet/type/iis_site.rb
MosesMendoza, line 10: Good point - we want to be able to `ensure => absent` here to remove the site

lib/puppet/type/iis_site.rb
```diff
@@ -0,0 +1,270 @@
...
+    end
+  end
+
+  newproperty(:physicalpath) do
+    desc 'The physical path to the IIS web site folder'
+    validate do |value|
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty physicalpath must be specified."
+      end
+      fail("File paths must be fully qualified, not '#{value}'") unless value =~ /^.:(\/|\\)/ or value =~ /^\/\/[^\/]+\/[^\/]+/
```
MosesMendoza, line 50: If I'm understanding correctly, in `/^.:(\/|\\)/` is it worth restricting the first character in case the user accidentally typos this (ex: `physicalpath => '@:/mysite'`)? May not be worth it if we don't get far.

lib/puppet/type/iis_site.rb
```diff
@@ -0,0 +1,270 @@
...
+    end
+    
+    if self[:serviceautostartprovidertype]
+      fail("Must specify serviceautostartprovidername as well as serviceautostartprovidertype") unless self[:serviceautostartprovidername]
+    end
+
+    provider.validate if provider.respond_to?(:validate)
+  end
+
+  def munge_boolean(value)
```
MosesMendoza, line 260: Instead of defining our own munge_boolean method, I think we could have properties that should be boolean use `Puppet::Property::Boolean` - 

ie

```
require 'puppet/property/boolean'
...
newproperty(:serviceautostart, :boolean => true, :parent => Puppet::Property::Boolean) do
```

and then the boolean properties don't need their own `munge` defined either (unless you want to do more than set boolean)

glennsarti, line 260: Is that a Puppet 4 thing or valid in Pup 3?

MosesMendoza, line 260: Added in puppet 3.3.0 - https://github.com/puppetlabs/puppet/commit/c06eb57923eb09b66f349202a2613bc91f1cb54c

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -0,0 +1,182 @@
+require_relative '../../../puppet_x/puppetlabs/iis/powershell_manager'
```
MosesMendoza: We may want to avoid `require_relative` (https://tickets.puppetlabs.com/browse/FM-5325), I believe to avoid breaking compat w/open source masters on distros. That is, if we want to worry about that scenario here. Instead we could use the (admittedly less concise) `require File.join(File.dirname(__FILE__), '../../../puppet_x/puppetlabs/iis/powershell_manager')` ?

lib/puppet/type/iis_site.rb
jpogran: @MosesMendoza I intially tried doing that, but I need to know when (https://github.com/puppetlabs/puppetlabs-iis/pull/3/files#diff-4e93d04350c2b0e0d149ceb808532cc2R25) it's not specified so as to not include the PS code to set it. If I use the suggested boolean stuff I get syntax errors when I check if it's present. Any suggestions?

ferventcoder: I once attempted this and ran into issues - and determined it only works for types built into Puppet for some reason. It could be I didn't have something set up properly - that was two years ago, so it's likely the require was missing.

ferventcoder: And booleans are either true or false, I don't know that they let you inspect a third state, which is undef

PR #6, opened 2016-11-30
Update Gemfile to install a current puppet version by default

DavidS:


DavidS:
Will supersede with proper modulesync update.

PR #8, opened 2016-11-30
WIP FM-5923 install iis as part of acceptance tests

tphoney:


tphoney:
@jpogran not sure what exact features we need installed ? 
@DavidS are you happy with the formatting ? did you want the manifest as a separate file, for end users ?

jpogran:
@tphoney Only the following are needed to run our module:

```
Web-WebServer
Web-Scripting-Tools
```

Technically, only `Web-WebServer` and `Web-Scripting-Tools` are absolutely needed, `Web-Mgmt-Console` is only useful for humans. The rest is either included in `Web-WebServer` or not needed for our purposes

tphoney:
@jpogran i made the suggested change

PR #9, opened 2016-12-01
copy module to correct location

tphoney:


tphoney:
@jpogran @DavidS , can i get a review

PR #11, opened 2016-12-02
test setups a simple iis site, metadata fixup

tphoney:


jpogran:
@tphoney you are encountering suitability issues with this because of line https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_site/webadministration.rb#L16. The machine you're using is windows2012, which has IIS 8.0 installed. This line allows 8 and 8.5, but https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/facter/iis_version.rb#L9 returns 8.0, and so we don't execute. I have this fixed locally, will put up a PR soon

tphoney:
@jpogran should of read this comment first before deep diving. anywho, fix made. tests run. though the resource is not idempotent yet.

PR #14, opened 2016-12-07
Add iis_application_pool type

DavidS:
This commit contains a bunch of testing infrastructure, mostly copied
over from puppetlabs-azure, to provide a nice experience when testing
native types and providers.

The tests now fail with "Could not run: iis_app_pool has no providers
and has not overridden 'instances'" on a puppet resource run.


tphoney:
looks good so far, just needs the provider :P 
also a boolean type would be good https://github.com/puppetlabs/puppetlabs-netscaler/blob/master/lib/puppet/property/netscaler_truthy.rb

tphoney:
state engine for flush from vanilla ice https://github.com/puppetlabs/vanilla_ice/blob/master/lib/puppet/provider/xe_ietf_interfaces/netconf.rb#L90-L108 @jpogran

lib/puppet/type/iis_app_pool.rb
```diff
@@ -0,0 +1,12 @@
+require 'puppet/parameter/boolean'
+require_relative '../../puppet_x/puppetlabs/iis/property/string'
+
+Puppet::Type.newtype(:iis_app_pool) do
```
jpogran: Pedantic naming discussion ahead, may not be useful right now:

Will we have` iis_app` or `iis_application` for MODULES-3050? I realize we have a convention for shortening names where possible, but I vote that `iis_applicationpool` or `iis_application_pool` is better at first glance to disambiguate than `iis_app_pool` next to `iis_app`.

DavidS: This is part of why manifests are good design artifacts. Generally the puppet idiom is to have more underscores. So the question is `iis_application_pool` vs `iis_app_pool`. Which one it is, I have no preference. I'll defer to the domain expert on that one.

In the end it is something a script can fix for us in the last moment.

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -1,18 +1,18 @@
...
 # of sites on a server increases, Get-Website will take longer and longer
 # to return. This will exponentially increase the total duration of your
 # puppet run
 
-Puppet::Type.type(:iis_site).provide(:webadministration) do
+Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
   desc "IIS Provider using the PowerShell WebAdministration module"
 
   require Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/puppetlabs/iis/powershell_manager'
   require Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/puppetlabs/iis/powershell_common'
```
jpogran: these two lines shouldn't be needed now that the psmanager stuff has been moved to a parent class

lib/puppet/provider/templates/webadministration/_getapppools.ps1.erb
```diff
@@ -0,0 +1,3 @@
+# TODO: this is JSON, PS v2 doesnt support it
+# PSCustomObject isnt present in v2
+Get-WebConfiguration -Filter '/system.applicationHost/applicationPools/add' | ConvertTo-Json -Depth 10
```
jpogran: I was trying to locate these ps1 templates inside each provider to avoid name collision or a massive folder. What do you think?

I advise doing what I did with iis_site and custom crafting the object to be passed to convert-json, as there is a severe performance penalty as well as a potential for errors when passing this directly to that cmdlet.

DavidS: just the provider name is not enough, as all providers will be called "webapplication".

lib/puppet/provider/iis_site/webadministration.rb
DavidS: ha, good catch.

lib/puppet/type/iis_app_pool.rb
tphoney: iis_application_pool FTW

lib/puppet_x/puppetlabs/iis/powershell_manager.rb
```diff
@@ -169,7 +169,7 @@ def self.create_event(name, manual_reset = false, initial_state = false)
       WAIT_TIMEOUT = 0x00000102
       WAIT_FAILED = 0xFFFFFFFF
 
-      def self.wait_on(wait_object, timeout_ms = 50)
+      def self.wait_on(wait_object, timeout_ms = 1000)
```
tphoney: does this need to be configurable ? (not as part of this PR to be clear)

tphoney:
can we get this merged soon :) 

eager to play

lib/puppet_x/puppetlabs/iis/powershell_manager.rb
jpogran: It actually shouldn't be messed with, the pain-point is the debug messages not the timing. IMHO the change should be removed from this PR as it was a red herring and we add another PR to discuss removing the debug message. The utility of the message stems from wanting to 'see' some output interactively so that the user can be sure something is happening, without this message there is no indication anything is happening.

lib/puppet/type/iis_app_pool.rb
jpogran: Same here. I vote this is changed now to prevent having to change lots of places later

jpogran:
I have some edits for the get application pools script, I can either push it to this or do another PR

lib/puppet_x/puppetlabs/iis/powershell_manager.rb
DavidS: Having the message every 50ms just spams the hell of the output. I felt one message per second is a much better tradeoff.

This change can easily be removed from the PR.

DavidS:
@jpogran feel free to push the changes on top of this - I'm still in full-blown prototype and rebase mode here, and can integrate your changes manually if necessary.

tphoney:
@DavidS this can be merged if we rename the app_pool provider and change the timeout back to @jpogran 's default

DavidS:
Updated w/ requested changes

PR #15, opened 2016-12-12
Extend wait time to reduce debug output

DavidS:
The original timeout of 50ms would create hundreds of lines of debug
output. This reduces that amount 20-fold, by printing a message every
second.


jpogran (approved):
This is fine for now, as I remembered we have to port the fixes from the powershell module in order to support 2016 here. The entire `powershell_manager` class is re-written to use named pipes, so most of this goes away.

PR #17, opened 2016-12-15
(MODULES-4219) Add Create/Destroy to IIS App Pool

jpogran:
This adds the create/update/destroy methods for the iis_application_pool
which handle the name, state, managedpipelinemode, and
managedruntimeversion properties.

This also adds acceptance tests covering iis_application_pool

Included tests:
 - Creation of app pool resource with default params
 - Creation of app pool resource with specified params
 - Creation of app pool resource with invalid params
 - Removal of app pool resource
 - Starting a stopped application pool

### DIFF
--- a/lib/puppet/provider/iis_application_pool/webadministration.rb
+++ b/lib/puppet/provider/iis_application_pool/webadministration.rb
@@ -10,26 +10,50 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
   commands :powershell => PuppetX::IIS::PowerShellCommon.powershell_path
 
   mk_resource_methods
-
+  
   def create
+    inst_cmd = "New-WebAppPool -Name \"#{@resource[:name]}\" -ErrorAction Stop"
+    result   = self.class.run(inst_cmd)
+    Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:exitcode] == 0
+    Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:errormessage].nil?
   end
 
   def update
+    inst_cmd = []
+
+    inst_cmd << self.class.ps_script_content('_setapppool', @resource)
+    
+    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state }.each do |property|
+      inst_cmd << "Invoke-AppCmd -ArgumentList 'set', 'apppool', #{@resource[:name]}, '/#{property.name.to_s}:#{property.value}';"
+    end
+    
+    if !@resource[:state].nil?
+      case @resource[:state]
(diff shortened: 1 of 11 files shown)


lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -29,6 +29,10 @@ def destroy
     Puppet.err "Error destroying website: #{result[:errormessage]}" unless result[:errormessage].nil?
     # return exists?
   end
+  
+  def flush
```
DavidS: https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_powershell.rb#L31

jpogran: yeah, I noticed that, when it tried to create the pool twice. both property_hash and original_values are empty so it tries to run create again. i thought i had remembered you saying each provider may have to override it to do special things in flush, so thought that this was the case here

DavidS: either there is something wrong with the hashes, or they are empty, because there are no properties yet. not sure which it is.

spec/acceptance/iis_application_pool_spec.rb
```diff
@@ -22,6 +22,17 @@
...
   end
+  
+  context "when setting props" do
+    it 'sets the state to stopped' do
+      apply_manifest(%{
+        iis_application_pool { 'test':
+          ensure => present,
+          state => 'stopped'
+        }
+      }, acceptable_exit_codes: (0...256))
```
DavidS: This is not testing anything. See https://github.com/puppetlabs/puppetlabs-iis/blob/master/spec/acceptance/iis_application_pool_spec.rb#L14 and https://github.com/puppetlabs/puppetlabs-iis/blob/master/spec/support/examples/idempotent_resource.rb for examples.

jpogran: Yep, I know, I was trying to have some test for the new properties being set, but still figuring out this syntax.

DavidS: :+1:

spec/acceptance/iis_application_pool_spec.rb
```diff
@@ -1,44 +1,143 @@
 require 'spec_helper_acceptance'
 
-describe 'iis_application_pool' do
+describe 'iis_application_pool', :focus => true do
```
DavidS: dev left overs

jpogran:
@wilson208 @ferventcoder please review

ThoughtCrhyme:
Unit tests look good to me.

PR #19, opened 2016-12-21
(MAINT) use beaker-windows to install features in spec_helper_acceptance

DavidS:
This removes the dependency on another module (puppet/windowsfeature),
and reduces "pollution" on the SUT.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -19,6 +19,7 @@ Gemfile:
   optional:
     ':system_tests':
       - gem: 'beaker-testmode_switcher'
+      - gem: 'beaker-windows'
 MAINTAINERS.md:
   maintainers:
     - "Puppet Windows Team `windows |at| puppet |dot| com`"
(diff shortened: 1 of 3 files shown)


DavidS:
This still fails with 

```
An error occurred in a `before(:suite)` hook.
Failure/Error: install_windows_feature_on(host, 'Web-WebServer')
NoMethodError:
  private method `exec' called for Host "":Serverspec::Type::Host
```

I guess we need to go to beaker-windows to get this fixed.

:-/

DavidS:
This depends on #20.

jpogran:
Couldn't get this working back then, left on drawing board. In the mean while, we implemented #30 to address windows 2008 support. While this ticket does it more the 'beaker' way, I don't think this is worth the extra time now to try to accomplish, when we have a working solution, and this is just for the acceptance test side. Closing for now

PR #22, opened 2017-01-05
(MODULES-4097) Sync travis.yml

hunner:
https://github.com/puppetlabs/modulesync_configs/pull/115


DavidS:
This doesn't look right: it drops the `--without system_tests`.

PR #26, opened 2017-01-24
(MODULES-4220) IIS Application Pool Extended properties

jpogran:


lib/puppet/type/iis_application_pool.rb
```diff
@@ -1,48 +1,395 @@
...
+  newproperty(:queue_length, :parent => PuppetX::PuppetLabs::IIS::Property::PositiveInteger) do
+    validate do |value|
+      super value
+      fail "#{self.name.to_s} should be greater than 10" unless value.to_i > 10
+      fail "#{self.name.to_s} should be less than or equal to 65535" unless value.to_i <= 65535
+    end
+  end
+
+  newproperty(:cpu_action, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
+    validate do |value|
```
DavidS: You can use `newvalue` here instead of custom code. See [docs](https://docs.puppet.com/guides/custom_types.html#specifying-allowed-values) and [coding examples](https://github.com/puppetlabs/puppetlabs-apt/blob/master/lib/puppet/type/apt_key.rb#L52). This will expose the information to the docs processor.

How stable are those values in MS' API, and how often will we have to release updates to keep up? Does the IIS API provide good error messages we could re-use instead of validating ourselves?

From a UX point of view: would it be helpful to be case-insensitive on the puppet side here?

spec/acceptance/iis_application_pool_spec.rb
```diff
@@ -50,8 +50,8 @@
         
         include_context 'with a puppet resource run'
         puppet_resource_should_show('ensure', 'present')
-        puppet_resource_should_show('managedpipelinemode', 'Classic')
-        puppet_resource_should_show('managedruntimeversion', 'v4.0')
+        puppet_resource_should_show('managed_pipeline_mode', 'Classic')
+        puppet_resource_should_show('managed_runtime_version', 'v4.0')
```
DavidS, line 54: puppet resource should show all the attributes, this should be testing for them.

lib/puppet/type/iis_application_pool.rb
```diff
@@ -1,48 +1,395 @@
...
+    # 00:05:00
+    # ($_ -eq '') -or
+    # (& {
+    #     ([ValidateRange(0, 86399)]$valueInSeconds = [TimeSpan]::Parse($_).TotalSeconds); $?
+    # })
+    validate do |value|
+      super value
+      fail "#{self.name.to_s} should be less than or equal to 86399" unless value.to_i <= 86399
+    end
+  end
```
DavidS: All of the properties should have short doc strings explaining what each is, mentioning allowed values (where `newvalues` is not sufficient), reference to the upstream parameter name (for searching), and the implied unit of the value.

Usually I copy the the summary line from upstream documentation under fair use.

DavidS (changes requested):

spec/acceptance/iis_application_pool_spec.rb
jpogran: WIP...haven't gotten there yet.

lib/puppet/type/iis_application_pool.rb
jpogran: WIP...haven't gotten there yet.

lib/puppet/type/iis_application_pool.rb
jpogran: TIL about `new value`.

No, the API is `appcmd` for the older stuff (non IISAdministration PowerShell Module) which outputs sometimes helpful messages like 'must be either true or false' or just an error number and generic 'this failed' message. The properties are fairly well documented in what they accept, it's a matter or writing each one.

spec/acceptance/iis_application_pool_spec.rb
jpogran: this has been addressed

glennsarti:
@jpogran Ran beaker acceptance locally.  Ready for merge

PR #27, opened 2017-01-27
[MODULES-3887] Implement iis_site flush method

wilson208:
Within this PR is the implementation of the flush method for the iis_site resource and accompanying acceptance tests to cover this. A number of small fixes were made as required.

**Implement flush for iis_site resource’s web administration provider:**
By reusing a lot of the same powershell code used when creating a new iis_site resource I implemented the `update` method within the iis_site webadministration provider. I abstracted the powershell Try-SetItemProperty to a new template file to avoid repetition.

**Acceptance Tests**
The following tests have been added for the iis_site resource and utility methods implemented to support these tests:
- Create a site with basic params
- Create a site with all params specified
  - Using W3C log format, logflags and logtruncatesize
  - Using NCSA log format and logperiod rather than logtruncatesize
  Change site state
  - From Stopped to Started
  - From Started to Stopped
  - From Started to Absent
- Ensure run fails with invalid params for:
  - logformat param
  - logperiod param
  Ensure can update existing, previously set param:
  - physicalpath param
  - enabledprotocols param
  - logflags param
- Utility methods added:
  - remove_all_sites
  - create_site(name, started, path)
  - create_path(path)

**Fixes:**
- iis_site webadministration provider:
  - Fix errors when no iis_site resources exist
  - Fix typo causing self.instances to return empty physicalpath's
  - In iis_site’s self.to_bool, handle case of empty value returns false in existing regex.
  - In iis_sites’s self.to_bool, return true or false symbols
  - Fix logproperties.ps1 template to correctly set array of flags as a comma-separated string when iis_site is being created or updated

- iis_site type:
  - Overwrite logflags insync? method to deal with differences in ordering of array

- Acceptance Test Support Files:
  - Standardise how we test for idempotency in the `idempotent_resource` shared example, so it is consistent with other modules
  - Escape values passed to `puppet_resource_should_show` to allow validation of path strings containing special characters (e.g. backslashes)

### DIFF
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -47,7 +47,7 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
   end
 
   def exists?
-    inst_cmd = "Get-Website -Name '\"#{@resource[:name]}\"'"
+    inst_cmd = "Get-Website -Name '#{@resource[:name]}'"
     result   = self.class.run(inst_cmd)
 
     resp = result[:stdout]
@@ -98,7 +98,7 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
   def self.prefetch(resources)
     sites = instances
     resources.keys.each do |site|
-      if provider = sites.find{ |s| s.name == site }
+      if !sites.nil? && provider = sites.find { |s| s.name == site }
         resources[site].provider = provider
       end
     end
@@ -107,33 +107,36 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
   def self.instances
     inst_cmd = ps_script_content('_getwebsites', @resource)
     result   = run(inst_cmd)
-    text     = result[:stdout]
 
-    site_json = JSON.parse(text)
+    return [] if result.nil? || result[:stdout].nil?
+
+    site_json = JSON.parse(result[:stdout])
+    return [] if site_json.nil?
+
     site_json = [site_json] if site_json.is_a?(Hash)
-    site_json.collect do |site|
+    return site_json.collect do |site|
       site_hash = {}
 
       site_hash[:ensure]               = site['state'].downcase
       site_hash[:name]                 = site['name']
       site_hash[:physicalpath]         = site['physicalpath']
       site_hash[:applicationpool]      = site['applicationpool']
-      site_hash[:serverautostart]      = to_bool(site['serverautostart'].downcase) unless site['serverautostart'].empty?
+      site_hash[:serverautostart]      = to_bool(site['serverautostart'])
       site_hash[:enabledprotocols]     = site['enabledprotocols']
       site_hash[:logpath]              = site['logpath']
       site_hash[:logperiod]            = site['logperiod']
       site_hash[:logtruncatesize]      = site['logtruncatesize']
-      site_hash[:loglocaltimerollover] = to_bool(site['loglocaltimerollover'].downcase) unless site['loglocaltimerollover'].empty?
+      site_hash[:loglocaltimerollover] = to_bool(site['loglocaltimerollover'])
       site_hash[:logformat]            = site['logformat']
-      site_hash[:logflags]             = site['logextfileflags']
+      site_hash[:logflags]             = site['logextfileflags'].split(/,\s*/).sort
 
       new(site_hash)
     end
   end
 
   def self.to_bool(value)
-    return true   if value == true   || value =~ (/(true|t|yes|y|1)$/i)
-    return false  if value == false  || value =~ (/(false|f|no|n|0)$/i)
+    return :true   if value == true   || value =~ (/(true|t|yes|y|1)$/i)
+    return :false  if value == false  || value =~ (/(^$|false|f|no|n|0)$/i)
     raise ArgumentError.new("invalid value for Boolean: \"#{value}\"")
   end
 end
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -22,6 +22,8 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
 
     cmd << self.class.ps_script_content('_newwebsite', @resource)
 
+    cmd << self.class.ps_script_content('trysetitemproperty', @resource)
+
     cmd << self.class.ps_script_content('generalproperties', @resource)
 
     cmd << self.class.ps_script_content('logproperties', @resource)
@@ -34,6 +36,29 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
 
     Puppet.err "Error creating website: #{result[:errormessage]}" unless result[:exitcode] == 0
     Puppet.err "Error creating website: #{result[:errormessage]}" unless result[:errormessage].nil?
+
+    return exists?
+  end
+
+  def update
+    cmd = []
+
+    cmd << self.class.ps_script_content('_setwebsite', @resource)
+
+    cmd << self.class.ps_script_content('trysetitemproperty', @resource)
+
+    cmd << self.class.ps_script_content('generalproperties', @resource)
+
+    cmd << self.class.ps_script_content('logproperties', @resource)
+
+    cmd << self.class.ps_script_content('serviceautostartprovider', @resource)
+
+    inst_cmd = cmd.join
+
+    result = self.class.run(inst_cmd)
+
+    Puppet.err "Error updating website: #{result[:errormessage]}" unless result[:exitcode] == 0
+    Puppet.err "Error updating website: #{result[:errormessage]}" unless result[:errormessage].nil?
 
     return exists?
   end
--- a/lib/puppet/provider/templates/webadministration/_setwebsite.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/_setwebsite.ps1.erb
@@ -1,0 +1,4 @@
+$resource = @{
+  name            = '<%= "#{resource[:name]}" %>'
+  ensure          = '<%= "#{resource[:ensure]}" %>'
+}
--- a/lib/puppet/type/iis_site.rb
+++ b/lib/puppet/type/iis_site.rb
@@ -206,24 +206,25 @@ Puppet::Type.newtype(:iis_site) do
     desc 'Specifies what W3C fields are logged in the IIS log file. This is only
       valid when :logformat is set to W3C. '
     validate do |value|
-      # if value.nil? or value.empty?
-      #   raise ArgumentError, "A non-empty defaultpage must be specified."
-      # end
-      unless value.kind_of?(Array) || value.kind_of?(String)
-        fail("Invalid value '#{value}'. Should be a string")
+      unless value.kind_of?(String)
+        fail("Invalid logflags value '#{value}'. Should be a string")
       end
       unless [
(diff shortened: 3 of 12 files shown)


lib/puppet/provider/iis_site/iisadministration.rb
```diff
@@ -1,23 +0,0 @@
-require 'pathname'
-
-Puppet::Type.type(:iis_site).provide(:iisadministration) do
```
jpogran: This file shouldn't be deleted

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -9,8 +9,8 @@
 Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
   desc "IIS Provider using the PowerShell WebAdministration module"
 
-  confine    :iis_version     => ['8.0','8.5']
-  confine    :operatingsystem => [:windows ]
+  confine :iis_version => ['8.0', '8.5']
```
jpogran: Is there a reason for the whitespace changes here?

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
     return exists?
   end
 
+  def update
+    Puppet.err 'UPDATE FUNCTION'
```
jpogran: TODO: remove

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
...
+          inst_cmd << "$site.LogFile.LogExtFileFlags='#{property.value}';"
+
+        else
+          inst_cmd << "$site.#{property.name.to_s}='#{property.value}';"
+      end
+    end
+    inst_cmd << '$serverManager.CommitChanges();'
+    inst_cmd = inst_cmd.join
+
+    Puppet.err('COMMAND: ' + inst_cmd)
```
jpogran: Puppet.debug works nicer here

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
...
+    result = self.class.run(inst_cmd)
+    Puppet.err('RESULT: ' + result[:stdout])
+
+    # Handle errors - had issue where when an issue occurred, errors were only output to stdout, not stderr
```
jpogran: This is powershell, not bash. We have to handle errors. See comment about erb above

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
     return exists?
   end
 
+  def update
+    Puppet.err 'UPDATE FUNCTION'
+
+    # Instantiate server manager object and get matching site using resource name
+    inst_cmd = []
+    inst_cmd << '[Void][Reflection.Assembly]::LoadWithPartialName("Microsoft.Web.Administration");'
```
jpogran: Would prefer this in an erb instead of a string array

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
...
+    inst_cmd << '$serverManager = New-Object Microsoft.Web.Administration.ServerManager;'
+    inst_cmd << "$site = $serverManager.Sites | where { $_.Name -eq '#{@resource[:name]}' };"
+    inst_cmd << "$app = $site.Applications | where { $_.Path -eq '/' };"
```
jpogran: These where clauses need to be tested on servers with more than a few sites. In prior testing with instances and create/delete, these methods get significantly slower as the number of sites go up

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -38,6 +38,59 @@ def create
...
+        when :logformat
+          inst_cmd << "$site.LogFile.LogFormat='#{property.value}';"
+        when :logflags
+          inst_cmd << "$site.LogFile.LogExtFileFlags='#{property.value}';"
+
+        else
+          inst_cmd << "$site.#{property.name.to_s}='#{property.value}';"
+      end
+    end
+    inst_cmd << '$serverManager.CommitChanges();'
```
jpogran: This needs a try/catch around it so we can present an error message

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -47,8 +100,11 @@ def destroy
   end
 
   def exists?
-    inst_cmd = "Get-Website -Name '\"#{@resource[:name]}\"'"
-    result   = self.class.run(inst_cmd)
+    Puppet.err('SITE Exists Called')
```
jpogran: All of these err calls either need to be changed to debug or removed

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -107,33 +163,36 @@ def self.prefetch(resources)
...
+          new(site_hash)
+        end
+      end
     end
   end
 
   def self.to_bool(value)
-    return true   if value == true   || value =~ (/(true|t|yes|y|1)$/i)
-    return false  if value == false  || value =~ (/(false|f|no|n|0)$/i)
+    return true if value == true || value =~ (/(true|t|yes|y|1)$/i)
```
jpogran: Its generally bad form to change whitespace without change to code, harder to read diffs

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -47,8 +61,8 @@ def destroy
   end
 
   def exists?
-    inst_cmd = "Get-Website -Name '\"#{@resource[:name]}\"'"
-    result   = self.class.run(inst_cmd)
+    inst_cmd = "Get-Website -Name #{@resource[:name]}"
```
hunner: This change breaks sites with spaces in the name

wilson208: Thanks @hunner, good catch!

wilson208:
@jpogran @glennsarti I have updated this PR with what we talked about last night about the performance of the powershell I was using. I have broken this down into smaller commits that detail everything I did. All spec & acceptance tests are now passing also.

lib/puppet/provider/templates/webadministration/_setwebsite.ps1.erb
```diff
@@ -0,0 +1,24 @@
+$resource = @{
+  name            = '<%= "#{resource[:name]}" %>'
+  ensure          = '<%= "#{resource[:ensure]}" %>'
+}
+
+function Try-SetItemProperty
```
jpogran: I would rather see this function pulled out of _newwebsite.ps1.erb into a general erb rather than duplicate it here

lib/puppet/type/iis_site.rb
```diff
@@ -206,25 +206,26 @@
...
+      unless value.kind_of?(String)
+        fail("Invalid logflags value '#{value}'. Should be a string")
       end
       unless [
-        'Date','Time','ClientIP','UserName','SiteName','ComputerName','ServerIP',
-        'Method','UriStem','UriQuery','HttpStatus','Win32Status','BytesSent',
-        'BytesRecv','TimeTaken','ServerPort','UserAgent','Cookie','Referer',
-        'ProtocolVersion','Host','HttpSubStatus'
-        ].include?(value)
+          'Date','Time','ClientIP','UserName','SiteName','ComputerName','ServerIP',
```
jpogran: I think you are adding tabs here? Only include needs to be moved

spec/support/examples/idempotent_resource.rb
```diff
@@ -1,9 +1,9 @@
 shared_examples 'an idempotent resource' do
   it 'should run without errors' do
-    apply_manifest(@manifest, :expect_changes => true)
+    apply_manifest(@manifest, :catch_failures => true)
```
jpogran, line 3: This is used by other tests. What does this change do to those?

spec/support/examples/idempotent_resource.rb
```diff
@@ -1,9 +1,9 @@
...
   end
 
   it 'should run a second time without changes' do
-    apply_manifest(@manifest, :expect_changes => false)
+    apply_manifest(@manifest, :catch_changes => true)
```
jpogran, line 7: This is used by other tests. What does this change do to those?

spec/support/matchers/puppet_resource_should_show.rb
```diff
@@ -4,8 +4,10 @@ def puppet_resource_should_show(property_name, value)
   it "should report the correct #{property_name} value" do
     regex = if value.nil?
               /(#{property_name})(\s*)(=>)(\s*)/
+            elsif(value.is_a?(Array))
+              /(#{property_name})(\s*)(=>)(\s*)(\[#{value.sort.map!{ |v| "'#{v}'" }.join(', ')}\])/i
             else
-              /(#{property_name})(\s*)(=>)(\s*)('#{value}'|#{value})/i
+              /(#{property_name})(\s*)(=>)(\s*)('#{Regexp.escape(value)}'|#{Regexp.escape(value)})/i
```
jpogran, line 10: This is used by other tests. What does this change do to those?

spec/support/utilities/iis_site.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+end
+
+def create_path(path)
+  on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& { New-Item -ItemType Directory -Force -Path '#{path}' }\"")
+end
+
+def remove_all_sites
+  on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& {Get-Website | Remove-Website}\"")
+end
```
jpogran: Nit: no newline at end of file

jpogran (changes requested):

spec/support/examples/idempotent_resource.rb
wilson208: This check for idempotency was broken previously and errors were not being raised, even if changes were made on the 2nd run of the given manifest.

spec/support/matchers/puppet_resource_should_show.rb
wilson208: This will not affect other tests at current. The issue was that for the first time in our tests we are ensuring a path property is set correctly and hence we have backslashes in the path that need to be escaped otherwise the regex will fail. This Regexp.escape function is specifically for escaping strings so they can be used in regular expressions and as such should not cause any issues.

spec/support/examples/idempotent_resource.rb
wilson208: This change was not necessary, however in a lot of our other modules tests we are using `:catch_changes` and `:catch_failures` rather than `:expect_changes` and `:expect_failures` and I was changing below from `:expect_changes => false` to `:catch_changes => true` and thought it best that I change this also for the sake of consistency.

wilson208: However this will not affect other tests.

lib/puppet/provider/templates/webadministration/_setwebsite.ps1.erb
wilson208: Have pulled this out now to trysetitemproperty.ps1.erb and pulling it in within the create and update methods.

lib/puppet/type/iis_site.rb
wilson208: So much for me 'cleaning up' my commits, missed this, sorted it now.

jpogran (approved):

jpogran:
👍

PR #29, opened 2017-02-01
(maint) Remove PowerShell 2 redirection test

jpogran:
This module does not rely on powershell redirection, and this test fails on PowerShell v2, so we have commented it out.


glennsarti (approved):
Pending CI PR

PR #30, opened 2017-02-02
(MODULES-4369) Use DISM to install iis

wilson208:
This replaces using puppet-windowsfeature module which is not compatible with Powershell 2.

This change uses puppetlabs-dism module to install IIS in the spec_helper_acceptance.rb file.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -29,13 +29,13 @@ RSpec.configure do |c|
       unless ENV['BEAKER_provision'] == 'no'
         shell("/bin/touch #{default['puppetpath']}/hiera.yaml")
         # install iis
-        on(agents, puppet("module install puppet/windowsfeature"))
+        on(agents, puppet("module install puppetlabs/dism"))
         pp=<<-EOS
-    $iis_features = ['Web-WebServer','Web-Scripting-Tools']
-
-    windowsfeature { $iis_features:
-      ensure => present,
-    }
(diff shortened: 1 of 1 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -29,13 +29,13 @@
       unless ENV['BEAKER_provision'] == 'no'
         shell("/bin/touch #{default['puppetpath']}/hiera.yaml")
         # install iis
-        on(agents, puppet("module install puppet/windowsfeature"))
+        on(agents, puppet("module install puppetlabs/dism"))
         pp=<<-EOS
-    $iis_features = ['Web-WebServer','Web-Scripting-Tools']
```
jpogran, line 34: We need both of these

spec/spec_helper_acceptance.rb
```diff
@@ -29,13 +29,13 @@
...
-
-    windowsfeature { $iis_features:
-      ensure => present,
-    }
+          dism { 'IIS-WebServer':
+            ensure    => present,
+            norestart => true,
+            all       => true
```
jpogran: While this will install the tools, it's overkill. We don't need the extra features, and it will add extra time to the tests

wilson208:
@jpogran Should be sorted now. Tested by running iis_application_pool tests and compared times with the old method of installing IIS using puppet-windowsfeature and the new DISM way and got similar times with bot, so I believe this is good to go if you are happy with it.

PR #31, opened 2017-02-03
(maint) Fix relative path for powershell provider

jpogran:
This fix ensures that the powershell base provider can be found during a puppet run, regardless of whether its using install puppet or bundle exec


glennsarti (approved):
Pending CI PR

PR #33, opened 2017-02-03
(MODULES-4370) Add PowerShell 2.0 and IIS 7.5 support

glennsarti:
Previously the module could not determine what scripts to execute as the
PowerShell version was unknown.  This commit adds a PowerShell version class and
method on the base IIS Provider to detect and cache the major PowerShell version.
This will be used in later methods to appropriately execute the correct scripts.

The IIS module uses JSON as the data transport between PowerShell and ruby.
However PowerShell 2.0 does not have a native JSON serializer.  This commit adds
the required PowerShell script to convert an object into a JSON string, and is
compatible with PowerShell 2.0.

Previously the module would not function on Windows 2008R2 which uses PowerShell
2.0 and IIS 7.5 by default.  This commit adds support by the following:

- Changes the usages of `[PSCustomObject]` to `New-Object ...` which will create
  appropriate object classes in PowerShell 2.0 and above.  Note that these are
  still objects of type PSCustomObject
- Added IIS version 7.5 to the confinement for the providers
- Added a JSON preparser as there are differences between the JSON generated by
  PowerShell 2.0 and 3.0+.  This method converts PowerShell 2.0 JSON objects
  into the 3.0+ equivalent
- Added a standard command header `add_powershell2_support` to all PowerShell
  invocations

Previously the type information for an object was being inserted into the JSON
documents.  While this does not break any existing functionality, future changes
to the module would have to deal with the extra information.  This commit sets
the NoTypeInformation to false and thereby stops all type information being
created.

### DIFF
--- a/lib/puppet/provider/iis_application_pool/webadministration.rb
+++ b/lib/puppet/provider/iis_application_pool/webadministration.rb
@@ -3,7 +3,7 @@ require File.join(File.dirname(__FILE__), '../../../puppet/provider/iis_powershe
 Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
   desc "IIS Application Pool provider using the PowerShell WebAdministration module"
 
-  confine    :iis_version     => ['8.0', '8.5']
+  confine    :iis_version     => ['7.5','8.0', '8.5']
   confine    :operatingsystem => [ :windows ]
   defaultfor :operatingsystem => :windows
 
@@ -61,11 +61,11 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
   def self.instances
     inst_cmd = ps_script_content('_getapppools', @resource)
     result   = run(inst_cmd)
-    text     = result[:stdout]
-    return [] if text.nil?
+    return [] if result.nil?
 
-    pool_json = JSON.parse(text)
-    pool_json = [pool_json] if pool_json.is_a?(Hash)
+    pool_json = self.parse_json_result(result[:stdout])
+    return [] if pool_json.nil?
+
     pool_json.collect do |pool|
       pool_hash = {}
 
--- a/lib/puppet/provider/iis_powershell.rb
+++ b/lib/puppet/provider/iis_powershell.rb
@@ -58,6 +58,52 @@ class Puppet::Provider::IIS_PowerShell < Puppet::Provider # rubocop:disable all
     PuppetX::IIS::PowerShellManager.instance("#{command(:powershell)} #{PuppetX::IIS::PowerShellCommon.powershell_args.join(' ')}")
   end
 
+  # do_not_use_cached_value is typically only used for testing. In normal usage
+  # the PowerShell version does not suddenly change during a Puppet run.
+  def self.ps_major_version(do_not_use_cached_value = false)
+    if @powershell_major_version.nil? || do_not_use_cached_value
+      version = self.powershell_version
+      @powershell_major_version = version.nil? ? nil : version.split('.').first.to_i
+    end
+    @powershell_major_version
+  end
+
+  PS_ONE_REG_PATH   = 'SOFTWARE\Microsoft\PowerShell\1\PowerShellEngine'
+  PS_THREE_REG_PATH = 'SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine'
+  PS_REG_KEY        = 'PowerShellVersion'
+
+  def self.powershell_version
+    Puppet::Util::Platform.windows? ? self.powershell_three_version || self.powershell_one_version : nil
+  end
+
+  def self.powershell_one_version
+    version = nil
+    reg_access = Win32::Registry::KEY_READ | 0x100
+
+    begin
+      Win32::Registry::HKEY_LOCAL_MACHINE.open(PS_ONE_REG_PATH, reg_access) do |reg|
+        version = reg[PS_REG_KEY]
+      end
+    rescue
+      version = nil
+    end
+    version
+  end
+
+  def self.powershell_three_version
+    version = nil
+    reg_access = Win32::Registry::KEY_READ | 0x100
+
+    begin
+      Win32::Registry::HKEY_LOCAL_MACHINE.open(PS_THREE_REG_PATH, reg_access) do |reg|
+        version = reg[PS_REG_KEY]
+      end
+    rescue
+      version = nil
+    end
+    version
+  end
+
   def self.ps_script_content(template, resource)
     @param_hash = resource
     template_path = File.expand_path('../templates', __FILE__)
--- a/lib/puppet/provider/iis_powershell.rb
+++ b/lib/puppet/provider/iis_powershell.rb
@@ -35,6 +35,16 @@ class Puppet::Provider::IIS_PowerShell < Puppet::Provider # rubocop:disable all
   end
 
   def self.run(command, check = false)
+    if self.ps_major_version == 2
+      # - PowerShell 2.0 does not support autoload of modules therefore we must explicitly add the WebAdministration module
+      # - Must change the current location to the be the IIS: provider
+      # - Add the ConvertTo-JSON command support
+      command = "Import-Module WebAdministration -ErrorAction Stop\n" +
+                "cd iis:\n" +
+                ps_script_content('json_1.7', @resource) + "\n" +
+                command
+    end
+
     Puppet.debug("COMMAND: #{command}")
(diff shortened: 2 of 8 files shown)


glennsarti:
There are currently no unit tests for this functionality.  Raised MODULES-4378 to add this.

eputnam (approved):

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -130,15 +146,18 @@ def self.prefetch(resources)
   end
 
   def self.instances
-    inst_cmd = ps_script_content('_getwebsites', @resource)
-    result   = run(inst_cmd)
+    cmd = []
+    cmd << self.add_powershell2_support if self.ps_major_version == 2
+
```
eputnam: whitespace on purpose?

lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -59,13 +67,18 @@ def self.prefetch(resources)
   end
 
   def self.instances
-    inst_cmd = ps_script_content('_getapppools', @resource)
+    cmd = []
+    cmd << self.add_powershell2_support if self.ps_major_version == 2
+
```
eputnam: in destroy, the `cmd` building is all together but in instances there's this space...

lib/puppet/provider/iis_powershell.rb
```diff
@@ -57,6 +58,50 @@ def self.ps_manager
     PuppetX::IIS::PowerShellManager.instance("#{command(:powershell)} #{PuppetX::IIS::PowerShellCommon.powershell_args.join(' ')}")
   end
 
+  def self.ps_major_version
+    if @ps_major_version.nil?
+      version = version = PuppetX::IIS::PowerShellVersion.version
```
eputnam: what is version = version = * doing?

glennsarti: Errr.... not much apparently.  I'll fix that.

lib/puppet/provider/iis_site/webadministration.rb
glennsarti: copy-pasta fails a-plenty

lib/puppet/provider/iis_application_pool/webadministration.rb
glennsarti: copy-pasta fails a-plenty

glennsarti: I was copying the existing code.  I will compress the extra lines.

glennsarti:
Rebased.

lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -12,39 +12,47 @@
   mk_resource_methods
   
   def create
-    inst_cmd = "New-WebAppPool -Name \"#{@resource[:name]}\" -ErrorAction Stop"
-    result   = self.class.run(inst_cmd)
+    cmd = []
```
jpogran: Would be good to not rename the main variable causing most of the diff to be green, when it's not needed for the purpose of the pr

lib/puppet/provider/iis_powershell.rb
```diff
@@ -1,4 +1,5 @@
 require 'pathname'
+require File.join(File.dirname(__FILE__), '../../puppet_x/puppetlabs/iis/powershell_version')
```
jpogran: This should be inside the IIS_Powershell class

lib/puppet/provider/iis_powershell.rb
```diff
@@ -57,6 +58,50 @@ def self.ps_manager
...
+    if @ps_major_version.nil?
+      version = PuppetX::IIS::PowerShellVersion.version
+      @ps_major_version = version.split('.').first.to_i
+    end
+    @ps_major_version
+  end
+
+  def self.add_powershell2_support
+    # - PowerShell 2.0 does not support autoload of modules therefore we must explicitly add the WebAdministration module
+    # - Must change the current location to the be the IIS: provider
```
jpogran: Why?

lib/puppet/provider/iis_powershell.rb
```diff
@@ -57,6 +58,50 @@ def self.ps_manager
...
+    # - Add the ConvertTo-JSON command support
+    ["Import-Module WebAdministration -ErrorAction Stop\ncd iis:\n",ps_script_content('json_1.7', @resource)]
```
jpogran: This is a lot in one place. Need to think about this

jpogran (changes requested):

lib/puppet/provider/iis_application_pool/webadministration.rb
glennsarti: I changed this one instance as all other methods used cmd. This was an outlier.  Would you prefer it in a separate commit?

lib/puppet/provider/iis_powershell.rb
glennsarti: This seems to be because we are running PowerShell in a custom host instead of interactively.  Without the `CD IIS:`,  Get-WebSite fails with a Null Exception.

Other examples:
https://forums.iis.net/t/1188540.aspx?Get+Website+Object+reference+not+set+to+an+instance+of+an+object+

https://forums.iis.net/t/1222290.aspx?Bug+report+NullReferenceException+when+using+the+WebAdministration+cmdlets+in+a+workflow

lib/puppet/provider/iis_powershell.rb
glennsarti: I will move the logic to this base class.

lib/puppet/provider/iis_powershell.rb
glennsarti: I'm a bit confused.  A lot of what in one place?

lib/puppet/provider/iis_powershell.rb
glennsarti: Fixed

glennsarti:
Added some simple integration tests for detecting the PowerShell version and the major version too.

lib/puppet/provider/iis_application_pool/webadministration.rb
jpogran: Just meant it wasn't relevant to this PR and made it hard to determine what changed. Nitpick.

lib/puppet/provider/iis_powershell.rb
jpogran: Interesting, TIL

lib/puppet/provider/iis_powershell.rb
jpogran: I meant it's a lot in one line, I didn't see the IIS provider part at first and it's hard to read.

lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -12,39 +12,47 @@
...
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:exitcode] == 0
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:errormessage].nil?
   end
 
   def update
-    inst_cmd = []
-
-    inst_cmd << self.class.ps_script_content('_setapppool', @resource)
+    cmd = []
+    cmd << self.add_powershell2_support if self.ps_major_version == 2
```
jpogran: Instead of remembering to add this line to each method, would it be better to add it to the `run` method? Inside run is where execution logic is, and is only one place to edit if we need to change this where we will have dozens of providers.

Something like:

```
cmd = []
if self.ps_major_version == 2
  cmd << self.add_powershell2_support
  cmd << command
else
  cmd = command
end
```

lib/puppet/provider/iis_powershell.rb
glennsarti: Ahh ok... I can make it look prettier

lib/puppet/provider/iis_application_pool/webadministration.rb
glennsarti: Hrmm....probably

glennsarti:
Rebased onto master.

lib/puppet/provider/iis_powershell.rb
glennsarti: Done

lib/puppet/provider/iis_application_pool/webadministration.rb
glennsarti: removed this change

jpogran (approved):

jpogran:
👍

PR #37, opened 2017-02-09
(MODULES-3832) Update PowerShell Manager Code

jpogran:
This updates the PowerShell manager code to the latest powershell module 2.1.0 which supports WMF 5.1 as well as some speed improvements.

This also paves the way for 2016 to be added to the jenkins pipeline. Note this does not add support to the iis module for 2016. That requires another set of work yet to be documented, the least of which means adding confines and providers for


glennsarti:
This has passed an adhoc run and is ready for merge.

PR #40, opened 2017-02-22
(MODULES-4434) IIS Feature management

jpogran:
This adds a type and provider to manage Windows feature installation on Windows 2008 and up

Note this code uses an array of the latest IIS features available to
install, but does not keep track of which subset is available in a given
IIS distribution. We could have kept track but since there are only a
handful of added features from 7.5 to 8.5, it was thought it would be
easier to have one array to check rather than keep a seperate list per
IIS version. In short, we defer to the tooling to tell us what feature is
present in which IIS version and only keep track of the larger list.

### DIFF
--- a/lib/puppet/provider/iis_feature/default.rb
+++ b/lib/puppet/provider/iis_feature/default.rb
@@ -1,0 +1,164 @@
+require File.join(File.dirname(__FILE__), '../../../puppet/provider/iis_powershell')
+
+Puppet::Type.type(:iis_feature).provide(:default, parent: Puppet::Provider::IIS_PowerShell) do
+  desc "IIS feature provider"
+
+  confine    :operatingsystem  => [ :windows ]
+  defaultfor :operatingsystem  => :windows
+
+  commands :powershell => PuppetX::IIS::PowerShellCommon.powershell_path
+
+  mk_resource_methods
+
+  def exists?
+    @property_hash[:ensure] == :present
+  end
+
+  def create
+    if iis_installable_features.includes?(@resource[:name])
+      raise Puppet::Error, "iis_feature can only be used to install IIS features. '#{resource[:name]}' is not an IIS feature"
+    end
+
+    if @resource[:include_management_tools] == true && self.class.is_windows2008 == true
+      raise Puppet::Error, 'include_management_tools can only be used with Windows 2012 and above'
+    end
+
+    cmd = []
+    cmd << "Import-Module ServerManager; Add-WindowsFeature #{resource[:name]}" if self.class.is_windows2008 == true
+    cmd << "Install-WindowsFeature #{resource[:name]}" if self.class.is_windows2008 == false
+    cmd << '-IncludeAllSubFeature' if @resource[:include_all_subfeatures] == true
+    cmd << '-Restart' if @resource[:restart] == true
+    cmd << "-Source #{resource['source']}" if @resource[:source] == false
+    cmd << '-IncludeManagementTools' if @resource[:include_management_tools] == true && self.class.is_windows2008 == false
+
+    Puppet.debug "Powershell create command is '#{cmd}'"
+    result = self.class.run(cmd.join)
+    Puppet.debug "Powershell create response was '#{result}'"
+  end
+
+  def destroy
+    if iis_installable_features.includes?(@resource[:name])
+      raise Puppet::Error, "iis_feature can only be used to uninstall IIS features. '#{resource[:name]}' is not an IIS feature"
+    end
+
+    cmd = []
+    cmd << "Import-Module ServerManager; Remove-WindowsFeature #{resource[:name]}" if self.class.is_windows2008 == true
+    cmd << "Uninstall-WindowsFeature #{resource[:name]}" if self.class.is_windows2008 == false
+    cmd << '-Restart' if @resource[:restart] == true
+
+    Puppet.debug "Powershell destroy command is '#{cmd}'"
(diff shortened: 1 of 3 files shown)


jpogran:
We discussed this as a team and resolved that it is acceptable to install and configure IIS in the same module because of precedent (apache, mysql, postgres...etc) and because it makes sense that this module knows how to do both, even though it smells like a separation of concerns issue. We are mitigating this concern by limiting this module loonily being able to install IIS features (in a commit to come).

lib/puppet/type/iis_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+    end
+  end
+
+  newproperty(:include_all_subfeatures, :boolean => true) do
+  end
+
+  newproperty(:restart, :boolean => true) do
```
hunner: Also a param.

lib/puppet/type/iis_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  newproperty(:include_management_tools, :boolean => true) do
```
hunner: Is this state that can be read from the system? If so, `self.instances` should set it, otherwise it should also be a param.

lib/puppet/type/iis_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  newproperty(:source, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
```
hunner: Also a param.

lib/puppet/type/iis_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+    desc "The unique name of the feature to manage."
+    validate do |value|
+      super value
+      if value.nil? or value.empty?
+        raise ArgumentError, "A non-empty #{self.name.to_s} must be specified."
+      end
+    end
+  end
+
+  newproperty(:include_all_subfeatures, :boolean => true) do
```
hunner: This should be a parameter if it cannot be read from the system (ie, this parameter is behavior, not state).

lib/puppet/provider/iis_feature/default.rb
```diff
@@ -0,0 +1,157 @@
...
+    cmd << 'Get-WindowsFeature | ConvertTo-Json -Depth 4'
+    
+    result = self.run(cmd.join)
+    
+    return [] if result.nil?
+
+    json = self.parse_json_result(result[:stdout])
+    return [] if json.nil?
+
+    return json.collect do |feature|
```
hunner: Ruby implicitly returns the last result in a method so this return isn't needed, but it's fine to have it. ruby-lint might warn about this.

hunner (changes requested):
Properties vs parameters

spec/acceptance/iis_feature.rb
```diff
@@ -0,0 +1,50 @@
...
+        before(:all) do
+          @result = on(default, puppet('resource', 'iis_feature', "Web-WebServer"))
+        end
+
+        include_context 'with a puppet resource run'
+        puppet_resource_should_show('ensure', 'present')
+      end
+
+      after(:all) do
+        # remove_app_pool(@pool_name)
```
glennsarti: nit. Not required.

spec/acceptance/iis_feature.rb
```diff
@@ -0,0 +1,50 @@
...
+          iis_feature { 'Foo':
+            ensure => 'present'
+          }
+          HERE
+        end
+
+        it_behaves_like 'a failing manifest'
+
+        after(:all) do
+          # remove_app_pool(@pool_name)
```
glennsarti: nit. Not Required

spec/acceptance/iis_feature.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper_acceptance'
+
+describe 'iis_feature' do
+  context 'when managing features' do
+    context 'with default parameters' do
+      before(:all) do
+        @manifest  = <<-HERE
```
glennsarti, line 7: I'm not sure this test is useful as pre-suite already installs the WebServer feature.  Perhaps we need to pick a non-default installed feature to test on.

Also as part of before(:all) may need to ensure that it does not exist before we test that we can install it.  Otherwise we're just testing two idempotent puppet runs, not an install puppet run and then and idempotent run.

lib/puppet/provider/iis_feature/default.rb
```diff
@@ -0,0 +1,157 @@
...
+  def self.os_major_version
+    if @os_major_version.nil?
+      version = Facter.value(:kernelmajversion)
+      @os_major_version = version.nil? ? nil : version
+    end
+    @os_major_version
+  end
+
+  def iis_installable_features
+    iis85 = [
```
glennsarti: Should there be `iis80` and `iis75` variables here too?

jpogran: Updated commit and PR message, commented code with the reason

hunner (approved):

spec/acceptance/iis_feature.rb
glennsarti: This has been changed to `Web-Asp-Net45`

glennsarti (approved):

PR #43, opened 2017-03-01
(maint) Fix iis_application_pool for updating after create

hunner:
After create, the provider's @property_hash still is empty so the update
function (called from flush) does not know what the state of the system
is any more.

This change re-fetches the provider instance and assigns it to the
resource. The transaction harness then calls update on the *new*
provider which has a populated @property_hash for correctly updating
properties.

### DIFF
--- a/lib/puppet/provider/iis_application_pool/webadministration.rb
+++ b/lib/puppet/provider/iis_application_pool/webadministration.rb
@@ -24,6 +24,7 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
     result   = self.class.run(inst_cmd)
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:exitcode] == 0
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:errormessage].nil?
+    @resource.provider = self.class.instances.find{|prov| prov.name == @resource[:name]}
     @resource[:ensure] = :present
   end
 
@@ -32,7 +33,9 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
     inst_cmd = []
     inst_cmd << self.class.ps_script_content('_setapppool', @resource)
 
-    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state }.each do |property|
+    @resource.properties.each do |property|
+      next if property.name == :ensure or property.name == :state
+      next if property.value != @property_hash[property.name]
       property_name = iis_properties[property.name.to_s]
       Puppet.debug "Changing #{property_name} to #{property.value}"
       inst_cmd << "Invoke-AppCmd -ArgumentList 'set', 'apppool', #{@resource[:name]}, '/#{property_name}:#{property.value}';"


lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -24,6 +24,7 @@ def create
     result   = self.class.run(inst_cmd)
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:exitcode] == 0
     Puppet.err "Error creating apppool: #{result[:errormessage]}" unless result[:errormessage].nil?
+    @resource.provider = self.class.instances.find{|prov| prov.name == @resource[:name]}
```
jpogran, line 27: Considering that calls to Get-Website are extremely expensive what cost does this have running again without filtering by name in the powershell side?

hunner: That was my comment when screen sharing; that this will run once per create and that could be slow, but I interpreted the answer as "don't worry about that." I have no idea how to make it faster... does "filtering" make it run faster?

jpogran: So in order to do what I suggest, we would have to modify this method to accept a variable to inject into the erb: https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_powershell.rb#L149-L154. Is that possible?

hunner: `puppet resource iis_application_pool <name>` when 4 pools exist takes 11 seconds
`puppet resource iis_application_pool <name>` when 150ish pools exist takes 12 seconds
`puppet resource iis_application_pool <name>` when 300ish pools exist takes 13 seconds
`puppet resource iis_application_pool <name> ensure=present` when 4 pools exist takes 15 seconds
`puppet resource iis_application_pool <name> ensure=present` when 150ish pools exist takes 17 seconds
`puppet resource iis_application_pool <name> ensure=present` when 300ish pools exist takes 20 seconds

So it does take a few seconds longer to create as more pools exist, but only 5 seconds longer at 300 pools than 3 pools.

hunner: Creating a pool when 300ish pools exist without this code takes 17 seconds, so adds up to +3 seconds per create with tons of pools.

hunner: I didn't find any way to re-fetch the pool and be anywhere near as fast, though I did find #45 to cut the current time to create / destroy in half. Results:
```
time to create 305 with modified exists? Notice: Applied catalog in 247.87 seconds
time to destroy 305 with standard code Notice: Applied catalog in 577.50 seconds
time to destroy 305 with modified exists? Notice: Applied catalog in 257.48 seconds
time to create 305 with modified create Notice: Applied catalog in 1283.89 seconds
time to create 305 with batched create and ps filtering Notice: Applied catalog in 1382.19 seconds
```

glennsarti (approved):

hunner:
Another thought about performance I had on this was that re-fetching the state could allow all property setters to be skipped, and if the users set 20 properties on every pool then it could be faster in cases of many properties but slower in cases of few properties.

In any case, this isn't a cut-and-dried case of functionality without performance impacts, so I'm going to close and move ahead with what I have.

PR #44, opened 2017-03-02
(MODULES-4463) Remove the use of Get-Website for Server 2008 compatibility

glennsarti:
Work in progress

### DIFF
--- a/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
@@ -13,10 +13,10 @@ $createParams = @{
   ErrorAction     = 'Stop'
 }
 
-# If there are no other websites, specify the Id
-# if (-not (Get-Website)){
-#   $createParams['Id'] = 1
-# }
+# If there are no other websites, specify the Id, otherwise an Index Out of Range error can be thrown
+If ((Get-ChildItem 'IIS:\sites' | Measure-Object).Count -eq 0) {
+  $createParams['Id'] = 1
+}
 
 # create website
 # dont set applicationpool if it doesnt exist
--- a/lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
@@ -1,9 +1,48 @@
+$iis_version = '<%= Facter.value(:iis_version) %>'
+
+# Convert the string representation of an Enum value into an Int32 safely
+Function ConvertTo-EnumInt32 {
+  param(
+    [Type]$EnumType,
+    [String]$Name,
+    [String]$Value
+  )
+
+  $EnumValue = ([Enum]::GetNames($EnumType)) | ForEach-Object {
+    if ($_.ToUpper() -eq $Value.ToUpper()) {
+      return [Int32]($EnumType::$_)
+    }
+  }
+  if ($EnumValue -eq $null) {
+    Throw "$Value is not a valid value for $Name"
+  } else {
+    Return $EnumValue
+  }
+}
+
+if ($iis_version -eq '7.5') {
+  # In IIS 7.5 the Enums are saved using the Int32 value, whereas in later versions, the text value can be used
+  # Load the WebAdministration DLL so we can convert string values into their Int32 enum equivalent
+
+  # Microsoft.Web.Administration namespace - https://msdn.microsoft.com/en-us/library/microsoft.web.administration(v=vs.90).aspx
+  # Log Format - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.logformat(v=vs.90).aspx
+  # LogExtFileFlags - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.logextfileflags(v=vs.90).aspx
+  # LogPeriod - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.loggingrolloverperiod(v=vs.90).aspx
+
+  [System.Reflection.Assembly]::LoadWithPartialName("Microsoft.Web.Administration") | Out-Null
+}
+
 <% if !resource[:logformat].nil? -%>
 $logParams = @{
   Path  = "IIS:\Sites\$($resource.name)"
   Name  = 'LogFile.logFormat'
   Value = '<%= "#{resource[:logformat]}" %>'
 }
+
+if ($iis_version -eq '7.5') {
+  $logParams.Value = ConvertTo-EnumInt32 -EnumType ([Microsoft.Web.Administration.LogFormat]) -Value $logParams.Value -Name 'logformat'
+}
+
 Try-SetItemProperty @logParams
 <% end -%>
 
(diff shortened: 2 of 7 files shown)


lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
```diff
@@ -1,9 +1,44 @@
...
+}
+# Microsoft.Web.Administration namespace - https://msdn.microsoft.com/en-us/library/microsoft.web.administration(v=vs.90).aspx
+# Log Format - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.logformat(v=vs.90).aspx
+# LogExtFileFlags - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.logextfileflags(v=vs.90).aspx
+# LogPeriod - https://msdn.microsoft.com/en-us/library/microsoft.web.administration.loggingrolloverperiod(v=vs.90).aspx
+
+if ($iis_version -eq '7.5') {
+  # In IIS 7.5 the Enums are saved using the Int32 value, whereas in later versions, the text value can be used
+  # Load the WebAdministration DLL so we can convert string values into their Int32 enum equivalent
+  [System.Reflection.Assembly]::LoadWithPartialName("Microsoft.Web.Administration") | Out-Null
```
jpogran: The assembly load should be moved into the function `ConvertStringTo-EnumInt32`, it was initially confusing to see it in one block and not the next further down.

This will also silently fail if not present. Worth to check and throw error?

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
```diff
@@ -1,9 +1,44 @@
+$iis_version = '<%= Facter.value(:iis_version) %>'
+
+$W3CValue = 'W3C'
+$MaxSizeValue = 'MaxSize'
```
jpogran: I don't understand the variable use here, when it is so far away from it's usage and it's hardcoded in either case

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
```diff
@@ -1,9 +1,44 @@
+$iis_version = '<%= Facter.value(:iis_version) %>'
+
+$W3CValue = 'W3C'
```
jpogran: I would rather this variable be closer to usage

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
```diff
@@ -1,9 +1,44 @@
...
+$MaxSizeValue = 'MaxSize'
+
+# Convert the string representation of an Enum value into an Int32 safely
+Function ConvertStringTo-EnumInt32($EnumType, $Value, $ValueName) {
```
jpogran: While syntactically correct, this is improper PowerShell function declaration. Please use a proper param statement:

```
function ConvertStringTo-EnumInt32 {
  param($EnumType, $Value, $ValueName)
}
```

jpogran: Would rather `ConvertStringTo-EnumInt32` be `ConvertTo-EnumInt32`, it breaks PowerShell verb-noun naming conventions and doesn't add any addition info, that couldn't be better expressed in a attributed param statement:

```
function ConvertTo-EnumInt32 {
  param(
    [string]$EnumType,
    [string]$Value,
    [string]$ValueName
    )
}
```

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
```diff
@@ -1,9 +1,44 @@
...
+  $EnumValue = ([Enum]::GetNames($EnumType)) | ForEach-Object {
+  if ($_.ToUpper() -eq $Value.ToUpper()) { return [Int32]($EnumType::$_) }
```
jpogran: Please line break this and indent

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb
jpogran: This parameter statement makes more sense to me. I was confused by `Value` being specified first, my first instinct is to specify the name of a thing and then it's value.

```
function ConvertTo-EnumInt32 {
  param(
    [string]$EnumType,
    [string]$Name, # formerly $Valuename
    [string]$Value 
    )
}
```

If you want further disambiguation, I could see it as `EnumName`, but I think that's redundant

Then use it:

```
ConvertStringTo-EnumInt32 -EnumType ([Microsoft.Web.Administration.LogFormat]) -Name 'logformat' -Value $logParams.Value

**Review by jpogran — CHANGES_REQUESTED** [review]

File: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb (hunk 1 above)
Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: I don't think it's appropriate to add the DLL load in there.  That function is entirely generic and can work on any Enum.  Where the DLL load is specific to IIS 7.5 only.

File: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb (hunk 3 above)
Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: Moved

File: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb (hunk 2 above)
Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: Moved

File: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb (hunk 4 above)
Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: Updated

Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: Updated. I've left it as EnumType as it's an Object of type System.Type.  I would rather not use the var name `$Enum` as it can get confusing with the keyword Enum and the Type System.Enum (`[Enum]`)

File: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb (hunk 5 above)
Comment (glennsarti) [reply lib/puppet/provider/templates/webadministration/logproperties.ps1.erb]: Updated.

File: lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb [hunk 6]
```diff
@@ -13,10 +13,8 @@ $createParams = @{
   ErrorAction     = 'Stop'
 }
 
-# If there are no other websites, specify the Id
-# if (-not (Get-Website)){
-#   $createParams['Id'] = 1
-# }
+# If there are no other websites, specify the Id, otherwise an Index Out of Range error can be thrown
+If ((Get-ChildItem 'IIS:\sites' | Measure-Object).Count -eq 0) { $createParams['Id'] = 1 }
```
jpogran: Please more than one line here

glennsarti: Updated.

jpogran (approved):

PR #49, opened 2017-03-10
(MODULES-3050) Add iis_application

hunner:


spec/support/utilities/iis_application.rb
```diff
@@ -0,0 +1,20 @@
+def has_app(app_name)
+  !(on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& {Get-WebApplication -Name #{app_name}}\"").stdout =~ /Started/i).nil?
```
jpogran: These should use the [helper](https://github.com/puppetlabs/puppetlabs-iis/blob/master/spec/support/utilities/utility_helper.rb) so it can work on windows2008

spec/support/utilities/iis_application.rb
```diff
@@ -0,0 +1,20 @@
...
+def remove_app(app_name)
+  on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& {Remove-WebApplication -Name #{app_name}}\"") if has_app(app_name)
+end
+
+def create_virtual_directory(site,name,directory)
+  on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& {New-WebVirtualDirectory -Site #{site} -Name #{name} -physicalPath #{directory} -Force -ErrorAction Stop}\"")
+end
+
+def create_selfsigned_cert(dnsname)
+  result = on(default, "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile -executionpolicy bypass -command \"& {(New-SelfSignedCertificate -DnsName '#{dnsname}' -CertStoreLocation cert:Localmachine\\My).Thumbprint}\"")
```
jpogran: This will break on windows 2008. `New-SelfSignedCertificate` only came into existence on 2012. This is one of the issues I called out when we discussed adding ssl to the acceptance tests. We now have to add functions to create self-signed on 2008

lib/puppet/provider/iis_application/webadministration.rb
```diff
@@ -0,0 +1,129 @@
+require 'puppet/provider/iis_powershell'
+
+Puppet::Type.type(:iis_application).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
+  desc "IIS Application provider using the PowerShell WebAdministration module"
+
+  confine    :iis_version     => ['8.0', '8.5']
```
jpogran, line 6: This prevents running on 7.5, which prevents windows2008. This means we will have to come back to this after the v0.1 release

lib/puppet/provider/iis_application/webadministration.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+  end
+
+  private
+
+  def check_paths
+    if @resource[:physicalpath] and ! File.exists?(@resource[:physicalpath])
+      fail "physicalpath doesn't exist: #{@resource[:physicalpath]}"
+    end
+    #XXX How do I check for IIS:\ path existence without shelling out to PS?
```
jpogran, line 125: You will have to go to PS, since IIS:\ is a PSProvider. Alternately you can go down to the raw xml. IMHO, go the PowerShell route

lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -27,7 +27,7 @@ def create
   def update
     Puppet.debug "Updating #{@resource[:name]}"
     inst_cmd = []
-    inst_cmd << self.class.ps_script_content('_setapppool', @resource)
+    inst_cmd << self.class.ps_script_content('invoke-appcmd', @resource)
```
jpogran: Nit: this along with the iis_site edits should probably be in separate commits.

jpogran (changes requested):

spec/support/utilities/iis_application.rb
hunner: We could conditionalize the ssl test to only run on 2012 if you'd like. I'd rather do that than not have any testing.

lib/puppet/provider/iis_application/webadministration.rb
hunner: This is just a nice-to-have so I'm going to skip it for now.

glennsarti (changes requested):
* Commit message is missing details
* Missing spec tests for the type and provider?

glennsarti (approved):

jpogran (approved):

PR #50, opened 2017-03-10
(MODULES-4484) Fix bug in ConvertTo-JSON on Windows 2008

glennsarti:
Previously, when a PowerShell command was sent to debug output it contained all
of the PowerShell code to support the ConvertTo-JSON command however this meant
the actualy command that was issued was lost if a sea of messages.  This commit
emits the debug message before all of the PowerShell support code is added.

Previously the ConverTo-JSON code was not serialising the bindings array
correctly.  Upon further investigation it was found that it was failing for all
single element arrays.  It was also found that if an object was an empty string
or empty array, the resultant JSON would say NULL instead of an empty object.
This behaviour is unfortunately inside the ConvertTo-XML cmdlet and could not
be rectified on the PowerShell side without also writing our own XML converter
too.

This commit;
- Adds a note to the JSON parser to be wary of nil values entering the result
  objects.  This can cause puppet to think properties have changed values or
  throw errors about operations on Nil classes
- In the iis_site webadministration provider, parse the result from PowerShell
  and convert any known String types that are nil to an empty string
- In the json converter:
  - Output the type information in the XML conversion as that's required by the
    XSLT to determine if an object is an array or a hashtable, previously it
    assumed all collections were a hashtable
  - Short circuit conversion if the input is a $null.  Otherwise the resultant
    JSON returned `{"":null}` instead of empty/blank
  - Previously the XLST would only consider collections with 2 or more elements
    as an array.  However this led to collections with only one element being
    mangled in the JSON output.  This commit removes the count restriction
  - Because the type information is being generated in the XML conversion, it
    subsequently needs to be removed via the XSL as Puppet does not know how to
    deal with the information.  This commit comments out the XSLT which would
    output the Type

### DIFF
--- a/lib/puppet/provider/iis_powershell.rb
+++ b/lib/puppet/provider/iis_powershell.rb
@@ -35,6 +35,8 @@ class Puppet::Provider::IIS_PowerShell < Puppet::Provider # rubocop:disable all
   end
 
   def self.run(command, check = false)
+    Puppet.debug("COMMAND: #{command}")
+
     if self.ps_major_version == 2
       # - PowerShell 2.0 does not support autoload of modules therefore we must explicitly add the WebAdministration module
       # - Must change the current location to the be the IIS: provider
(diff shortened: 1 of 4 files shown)


glennsarti:
Blocked on PR #47 and pending release.

glennsarti:
This has passed an adhoc run.  Works as expected on OSes other than 2008 too.

jpogran (approved):
Approved based of peer review and adhoc CI passing

PR #57, opened 2017-03-27
(MODULES-4464) Fix Set-ItemProperty use

jpogran:
This fixes using Set-ItemProperty. Previously a single backslash was used, should be a double


glennsarti:
Nit - The commit message says `Remove Appcmd fix` but doesn't seem to match the changes that are made

jpogran:
@glennsarti changed for better nit readability :)

PR #61, opened 2017-03-29
[maint] Fix development group naming in .sync.yml

wilson208:
This current .sync.yml produces an invalid Gemfile. The development group is specified in the Gemfile with a symbol and as such needs to be prefixed with a colon, like the :system_tests group below.


tphoney:
👍  lines up with what is in IIS

PR #62, opened 2017-04-03
Correct method name

binford2k:


jpogran:
@binford2k could you squash these two commits and prepend the commit message with (maint)?

jpogran (changes requested):
Squash commits and use (maint) message and this is good to go

jpogran (approved):

PR #63, opened 2017-04-10
(maint) Handle non-https bindings better; add docs

hunner:
The docs for sslflags and the two certificate attributes for a binding
describe the possible values and the fact that they are only needed when
protocol is the 'https' value.


hunner:
This passed ad-hoc CI tests

jpogran (approved):

PR #66, opened 2017-04-21
(MODULES-3053) IIS virtual directory

jpogran:
This PR adds a type and provider is added to manage IIS virtual
directories.


glennsarti:
Running through adhoc.

PR #72, opened 2017-05-05
(MODULES-4564) Document the iis_application resource

glennsarti:
Previously in MODULES-3050 the iis_application resource was added however there
was no documentation.  This commit adds the documentation to README.MD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 4. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
     * [Facts](#facts)
     * [Types/Providers](#types/providers)
+        * [iis_application](#iis_application)
         * [iis_application_pool](#iis_application_pool)
         * [iis_feature](#iis_feature)
         * [iis_site](#iis_site)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,7 @@
 4. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
     * [Facts](#facts)
     * [Types/Providers](#types/providers)
```
jbondpdx, line 10: Types and providers rather than Types/Providers

Usually we list types and providers separately, though I'm fine with listing them together if that makes the most sense. Are providers to be added later? If there's a provider for each type, we could include that information in the type descriptions instead. Feel free to ping me to chat about how to handle this bit! :-)

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
 
 ### Types/Providers
```
jbondpdx, line 69: Also here: Types and providers will need to match the link in the ToC

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
...
 
+* [iis_application](#iis_application)
 * [iis_application_pool](#iis_application_pool)
 * [iis_feature](#iis_feature)
 * [iis_site](#iis_site)
 
 Here, include a complete list of your module's classes, types, providers, facts, along with the parameters for each. Users refer to this section (thus the name "Reference") to find specific details; most users don't read it per se.
```
jbondpdx: If this is left in as a placeholder/reminder because there's more work to do, no worries. If not needed though, probably best to remove it so it doesn't accidentally get left in.

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
...
 
+### iis_application
+
+Allows creation of a new IIS application.
+
+#### `ensure`
+
+Must be either `present` or `absent`. Present will ensure the application pool is created.
```
jbondpdx: present and absent are strings, so single quotes. While you're here, please change "will ensure" to "ensures"?

I assume this has a default of 'present'? This and the other parameters in this section should have default values, if they exist.

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
...
+
+The name of the site for this IIS Web Application.
+
+#### `physicalpath`
+
+The physical path to the IIS web application folder.  This path must be fully qualified.
+
+#### `applicationpool`
+
+The name of an application pool for this IIS Web Application
```
jbondpdx: needs a period

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
...
+
+#### `virtual_directory`
+
+The Web Virtual Directory to convert to a Web Application on create.
+
+#### `sslflags`
+
+The ssl settings for the application.  Valid options are an array of 'Ssl, 'SslRequireCert', 'SslNegotiateCert', 'Ssl128'.
```
jbondpdx: SSL capped, I believe

README.md
```diff
@@ -67,12 +68,78 @@ iis_site { 'minimal':
...
+
+##### Example
+
+Require the use SSL
```
jbondpdx: the use of SSL?

jbondpdx (changes requested):

README.md
glennsarti: I'm a bit confused with this. This PR is for adding the `iis_application` documenation.

README.md
glennsarti: May need to address that in a separate PR.

jbondpdx: I don't think it's a giant deal, so later is fine.

README.md
jbondpdx: I mean this paragraph (where I added the note), which is from the template and should be removed before the module is released:

> Here, include a complete list of your module's classes, types, providers, facts, along with the parameters for each. Users refer to this section (thus the name "Reference") to find specific details; most users don't read it per se.

PR #74, opened 2017-05-05
(MODULES-4378) Add unit tests for parse_json_result

glennsarti:
Previously in MODULES-4370 the parse_json_result method was added but tests were
not.  This commit adds unit tests for this method and includes PowerShell 2 vs 3
responses.


jpogran:
LGTM if it passes adhoc

glennsarti:
Given that these are unit tests and have passed Travis and Appveyor, do I need adhoc?

jpogran:
Good point

PR #81, opened 2017-06-14
Adds support for IIS 10.0

rdev5:
Confines in the following providers have been extended to support 10.0 and tested against IIS 10.0 running on Windows Server 2016 Standard.

- `iis_application_pool`
- `iis_site`
- `iis_application`

Also fixes incorrect IIS version reporting in `iis_version.rb`.

### DIFF
--- a/lib/facter/iis_version.rb
+++ b/lib/facter/iis_version.rb
@@ -12,7 +12,12 @@ Facter.add("iis_version") do
       HKLM.open(REG_PATH, ACCESS_TYPE) do |reg|
         iis_ver = reg[REG_KEY]
       end
-      iis_ver = iis_ver[8,3]
+     
+      if iis_ver.match(/^Version (\d+\.\d+)$/) then
+        iis_ver = $1
+      else
+        iis_ver = iis_ver[8,3] # Incorrectly reads 10.0 as 10.
+      end
     rescue
       iis_ver = ""
     end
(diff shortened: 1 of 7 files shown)


jpogran:
Hi @rdev5, thanks for the PR!

We were planning on adding IIS 10 and Windows 2016/Nano support by using the new [IISAdministration](https://technet.microsoft.com/en-us/itpro/powershell/windows/iisadministration/iisadministration) cmdlets in another provider. For example, [iis_site](https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_site/iisadministration.rb) had the beginnings of this idea started but not finished. We were looking at taking advantage of the speed and resource improvements of the new cmdlets for 2016, and thus confined the old WebAdmistration module stuff to the older IIS versions. This means more work as it means making new code to handle the differently named cmdlets and different workflows that the new module uses, but we feel the benefits outweigh the extra work.

If you want to work towards that I'd be happy to advise, but we're going to keep restricting 10 inside the webadministration provider.

The parameter validation additions are useful, but we would need to add a minimal spec test to validate the validation (no pun intended). You can see an example [here](https://github.com/puppetlabs/puppetlabs-iis/blob/master/spec/unit/puppet/type/iis_site_spec.rb#L11-L39). If you're not familiar with rspec I can add a commit for them if you want.

rdev5:
I'm familiar with rspec but it's been a while. If you could go ahead and add a commit, that would be great. Thanks!

glennsarti:
@rdev5 I'm closing this PR.

* Your IIS 10.0 work was pulled into #123.  I needed to make some more test fixes for it to pass
* The naming changes were already pulled in via #113 #120 

Thanks again for the PR!!

PR #84, opened 2017-06-29
(MODULES-5125) Enable update IIS properties

jpogran:
This commit enables updating the physicalpath and applicationpool of an IIS website, and adds tests to verify


glennsarti:
Verified manually

PR #87, opened 2017-07-18
(MODULES-5264) IIS : physicalpath regex doesn't match UNC paths

tkishel:
Prior to this commit, the regex used to validate physicalpath
failed to match a valid UNC path. (Slashes leaning right instead of left.)
This commit tips the slashes to the left.


tkishel:
I found code that assumes physicalpath can be tested with File.exists:

https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_virtual_directory/webadministration.rb#L25

I'll submit a new PR ...

PR #88, opened 2017-07-20
Modules 5264 physicalpath regex does not match unc paths v2

tkishel:
(MODULES-5264) IIS physicalpath regex doesn't match UNC paths  …

The first commit corrects the regex in the types:

  forward vs backslashes

The second commit:

  corrects tests of physicalpath in the providers

    remove File.exists tests of UNC paths

  provides a workaround for New-WebVirtualDirectory and Set-ItemProperty
  when setting physicalpath when the UNC path is not available

  corrects typo: in provider iis_virtual_directory

    "applicaiton"

  corrects :ensure instance attribute in iis_site provider

    "Notice: ... ensure changed '' to 'present'"

  normalizes arbitrarily unique code across providers

### DIFF
--- a/lib/puppet/provider/iis_application/webadministration.rb
+++ b/lib/puppet/provider/iis_application/webadministration.rb
@@ -11,11 +11,6 @@ Puppet::Type.type(:iis_application).provide(:webadministration, parent: Puppet::
 
   mk_resource_methods
 
-  def initialize(value={})
-    super(value)
-    @property_flush = {}
-  end
-
   def physicalpath=(value)
     @property_flush[:physicalpath] = value
   end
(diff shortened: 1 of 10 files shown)


jpogran:
Hi @tkishel, thanks for the PR! Besides working on making the tests work, a few suggestions:

We generally try to not move blocks of code around without reason, as it confuses the diff and makes it harder to figure out what was really changed. For example, the destroy method was moved down below update. Could you please restore the original order of the methods and retain your changes?

There is a lot of code formatting, or just re-organizing of adjacent lines of code. This also makes it hard to figure out what the actual fix is. You can submit another PR and term it a maintenance PR and address the formatting you desire there.

This PR does too much in that it changes iis_site, iis_application, and iis_virtual_directoy all in the same commit and the same PR. The ticket you reference is just for iis_site, we would normally break these out into separate PRs or at least separate commits.

I have some more comments in specific lines to add, but can't right at the moment. Thanks again for the hard work, and look forward to getting this merged in.

tkishel:
I know, I know ... I'm sorry about that, and I accidentally dropped an applicationpool parameter that's the cause of the majority of errors. I'm unable to run tests locally: so I am depending upon appveyor and travis post PR.

I moved the code around because each provider implemented arbitrarily different code that I needed to normalize to comprehend and verify that the changes were compatible. I will try to rebsubmit as separate PRs.

iis_site, iis_application, and iis_virtual_directory all use paths for which the regexs will fail when passed a UNC path. I've verified each path that allows a UNC via Microsoft's documentation.

tkishel:
Closing this in favor of another PR with the smallest viable footprint.

PR #89, opened 2017-07-21
(MODULES-5259) Accept both uppercase and lowercase thumbprints

hunner:


spec/acceptance/iis_application_spec.rb
```diff
@@ -108,7 +108,7 @@
                 'bindinginformation'   => '*:443:#{site_hostname}',
                 'protocol'             => 'https',
                 'certificatestorename' => 'MY',
-                'certificatehash'      => '#{thumbprint}',
+                'certificatehash'      => '#{thumbprint.downcase}',
```
glennsarti, line 111: Do we need an additional test case to prove that thumbprints are case insensitive or is this enough?  I wonder if the IIS spec says all thumbprints are lower

hunner: IIS accepts both upper and lower cases, but only returns one of them. IIRC it returns lowecase in the gui and uppercase in powershell.

Maybe a better fix would be to compare the passed thumbprint to both upper and lowercase versions of whatever IIS has, but just upcasing it was the easiest.

jpogran: Upper or lower case, the thumbprint is the same. This is good to go

PR #93, opened 2017-07-27
(MODULES-5169) iis_application_pool not idempotent

tkishel:
For certain parameters iis_application_pool was not idempotent. This commit fixes this by using the correct IIS names for the parameters when evaluating them for modification. It also now handles parameters that are hashes or arrays correctly.

### DIFF
--- a/lib/puppet/provider/iis_application_pool/webadministration.rb
+++ b/lib/puppet/provider/iis_application_pool/webadministration.rb
@@ -28,18 +28,17 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
     Puppet.debug "Updating #{@resource[:name]}"
     cmd = []
 
-    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state && rp.name != :restart_schedule }.each do |property|
+    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state}.each do |property|
       property_name = iis_properties[property.name.to_s]
       Puppet.debug "Changing #{property_name} to #{property.value}"
-      cmd << "Set-ItemProperty -Path 'IIS:\\AppPools\\#{@resource[:name]}' -Name '#{property_name}' -Value #{property.value};"
-    end
-
-    # TODO: Update not called when restart_schedule is changed to [].
-    property_name = iis_properties['restart_schedule']
-    Puppet.debug "Changing #{property_name} to #{@resource[:restart_schedule]}"
-    cmd << "Clear-ItemProperty -Path 'IIS:\\AppPools\\#{@resource[:name]}' -Name '#{property_name}';"
-    @resource[:restart_schedule].each do |restart_time|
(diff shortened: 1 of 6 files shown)


lib/puppet/provider/iis_application_pool/webadministration.rb
```diff
@@ -28,10 +28,17 @@ def update
     Puppet.debug "Updating #{@resource[:name]}"
     cmd = []
 
-    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state }.each do |property|
+    @resource.properties.select{|rp| rp.name != :ensure && rp.name != :state}.each do |property|
       property_name = iis_properties[property.name.to_s]
-      Puppet.debug "Changing #{property_name} to #{property.value}"
-      cmd << "Set-ItemProperty -Path 'IIS:\\AppPools\\#{@resource[:name]}' -Name '#{property_name}' -Value #{property.value};"
+      if @resource[:name].is_a?(Hash)
```
jpogran: @tkishel what do you think of this approach instead?

lib/puppet/provider/templates/webadministration/_getapppools.ps1.erb
```diff
@@ -35,24 +35,26 @@ Get-WebConfiguration -Filter '/system.applicationHost/applicationPools/add' | %
...
+    auto_shutdown_params              = [string]$_.failure.autoshutdownparams
+
+    disallow_overlapping_rotation      = [string]$_.recycling.disallowoverlappingrotation
+    disallow_rotation_on_config_change = [string]$_.recycling.disallowrotationonconfigchange
+    log_event_on_recycle               = [string]$_.recycling.logeventonrecycle
+    restart_memory_limit               = [string]$_.recycling.periodicrestart.memory
+    restart_private_memory_limit       = [string]$_.recycling.periodicrestart.privatememory
+    restart_requests_limit             = [string]$_.recycling.periodicrestart.requests
+    restart_time_limit                 = [string]$_.recycling.periodicrestart.time
+    restart_schedule                   = [string]$_.recycling.periodicrestart.schedule.collection.value -split ' '
```
jpogran: Not sure we should cast to [string] here when we're asking it to split on spaces

tkishel: Actually, would it be better to limit PowerShell to simple data collection and move the split into the calling ruby code? Otherwise, yes, the cast should be to an array.

lib/puppet/provider/iis_application_pool/webadministration.rb
tkishel: Yeah, abstracting elements based on type is better than hard-coding.
But it needs to be: `if property.value.is_a?(Array)` and `property.value.each do |item|`
Should I add a commit or rebase with these changes?

lib/puppet/provider/templates/webadministration/_getapppools.ps1.erb
tkishel: I moved the split logic to webadministration.rb.

jpogran:
I rebased this from master. Waiting on adhoc CI to finish. Once that is done I will squash commits and merge

PR #94, opened 2017-08-02
(MODULES-5381) managed_runtime_version should allow the 'No Managed Code' value

carlosfbteixeira:
When setting up .Net CLR version under Application Pool settings on IIS there is no allowed value in the lib\puppet\type\iis_application_pool.rb for the value 'No Managed Code'. 
Windows translates the value "" to 'No Managed Code' on the application pool settings level.

https://tickets.puppetlabs.com/browse/MODULES-5381

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,15 @@ iis_application_pool { 'complete_site_app_pool':
   state                   => 'Started'
 }
 
+#Application Pool No Managed Code .Net CLR Version set up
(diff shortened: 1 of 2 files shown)


jpogran:
HI @carlosfbteixeira thanks for the PR! I adjusted your commit message so it added a little more detail to the message so someone coming to this commit later would know what's going on.

PR #101, opened 2017-08-10
(MODULES-5429) Fix iis_site idempotency

jpogran:
This commit fixes the start and stop methods on the iis_site provider by
adding error catching to the IIS cmdlet invocations. It now properly fails
if Puppet is unable to start or stop a website.

Previously, when issuing a start or stop command that failed, the error
would be passed to STDOUT and not reconized as an error, so Puppet would
report the state had changed and return success. When run again, Puppet
would again miss the error and report the state has changed, causing a
never ending loop of report notifications.

Unfortunately the error returned by IIS is generic, so the reason for not
being able to start or stop the site is unknowable to Puppet during the
run. It is up to the user to figure out the conflict.

### DIFF
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -47,6 +47,8 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
 
     cmd << self.class.ps_script_content('serviceautostartprovider', @resource)
 
+
+
     inst_cmd = cmd.join
 
     result = self.class.run(inst_cmd)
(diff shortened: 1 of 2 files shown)


glennsarti:
Running through adhoc

glennsarti:
Passed adhoc

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -75,34 +77,24 @@ def exists?
 
   def start
     create if ! exists?
-    @resource[:ensure]  = 'started'
 
-    inst_cmd = "Start-Website -Name \"#{@resource[:name]}\""
+    inst_cmd = "Start-Website -Name \"#{@resource[:name]}\" -ErrorVariable foo;if($foo){ throw \"$($foo). Perhaps there is another website with this port or configuration setting\" }"
```
glennsarti: Is foo the best name here?

glennsarti: I'll rename it and push up the change.

glennsarti:
This will supersede some of the work in #98

PR #103, opened 2017-08-15
(MODULES-5465) convert version confine from fact to feature

tkishel:
Prior to this commit, if IIS is installed during a run via iis_feature,
other iis_* types fail with 'Could not find a suitable provider' errors.
This is caused by confine testing a fact.

This confine tests a feature, which can (see PUP-5985) be re-evaluated,
and updates logproperties.ps1.erb to use PowerShell instead of a fact.

### DIFF
--- a/lib/puppet/feature/iis_compatible_version.rb
+++ b/lib/puppet/feature/iis_compatible_version.rb
@@ -1,0 +1,46 @@
+require 'puppet/util/feature'
+
+def iis_compatible_versions
+  ['7.5', '8.0', '8.5', '10.0']
+end
+
+def iis_installed_version
+  Puppet.debug "Reading Registry for IIS Version"
+  installed_version = nil
+  begin
+    require 'win32/registry'
+    hklm        = Win32::Registry::HKEY_LOCAL_MACHINE
+    reg_path    = 'SOFTWARE\Microsoft\InetStp'
+    access_type = Win32::Registry::KEY_READ | 0x100
+    reg_key     = 'VersionString'
+    iis_version_text = ''
+    hklm.open(reg_path, access_type) do |reg|
+      iis_version_text = reg[reg_key]
+    end
+    if iis_version_text.match(/^Version (\d+\.\d+)$/)
+      installed_version = $1
+    end
+    Puppet.debug "IIS Version: #{installed_version}"
+  rescue Exception => e
+    installed_version = nil
+  end
+  installed_version
+end
+
+def iis_compatible_version_installed?
+  iis_compatible_versions.include? iis_installed_version
+end
+
+##
+
+Puppet.features.add(:iis_compatible_version) do
+  iis_compatible_version_installed?
+end
+
+# https://tickets.puppetlabs.com/browse/PUP-5985
+
+Puppet.features.send :meta_def, 'iis_compatible_version' do
+  name = :iis_compatible_version
+  @results[name] = iis_compatible_version_installed? unless @results[name]
+  @results[name]
+end
--- a/lib/puppet/provider/iis_application/webadministration.rb
+++ b/lib/puppet/provider/iis_application/webadministration.rb
@@ -3,7 +3,7 @@ require File.join(File.dirname(__FILE__), '../../../puppet/provider/iis_powershe
 Puppet::Type.type(:iis_application).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
   desc "IIS Application provider using the PowerShell WebAdministration module"
 
-  confine    :iis_version     => ['7.5', '8.0', '8.5', '10.0']
+  confine    :feature         => :iis_compatible_version
   confine    :operatingsystem => [ :windows ]
   defaultfor :operatingsystem => :windows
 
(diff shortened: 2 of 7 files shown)


ferventcoder:
Do features get reevaluated? I like it

tkishel:
Yes, according to this: 

https://docs.puppet.com/puppet/latest/provider_development.html#suitability

"Using custom features you can delay resource evaluation until the provider becomes suitable."

Except for this: PUP-5985. 

So I implemented the workaround in PUP-5985. I don't think this feature check (a registry read) is too expensive to be reevaluated during a run, but I'll leave that for others to consider.

ferventcoder:
Registry reads are cheap - for the Chocolatey resource, we used validate instead of confine.

ferventcoder:
I think if this is repeating some of what the fact does, it's good to keep the fact around, but implement the functionality under puppetx and point both this and the fact to that resource to reuse logic.

tkishel:
I didn't know you could validate() outside of attributes/params :) I see that now:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L2292

Though its not documented in:

https://docs.puppet.com/puppet/latest/provider_development.html

I would like to eliminate the duplicate registry code in:

lib/facter/iis_version.rb
lib/feature/iis_compatible_version_installed.rb

But think its less of an issue in:

lib/puppet/provider/templates/webadministration/logproperties.ps1.erb

Thanks for the advice!

lib/puppet/feature/iis_compatible_version_installed.rb
```diff
@@ -0,0 +1,43 @@
+require 'puppet/util/feature'
+
+# Consider joining MajorVersion + MinorVersion instead of using VersionString
+
+def read_registry_for_iis_compatible_version
```
Iristyle: Since this returns a boolean, should rename method to `iis_compatible?` or similar.

https://github.com/bbatsov/ruby-style-guide#bool-methods-prefix

lib/puppet/feature/iis_compatible_version_installed.rb
```diff
@@ -0,0 +1,43 @@
...
+  installed_version = '0.0'
+
+  Puppet.debug "Reading Registry for IIS Version"
+
+  begin
+    require 'win32/registry'
+
+    f_HKLM        = Win32::Registry::HKEY_LOCAL_MACHINE
+    f_REG_PATH    = 'SOFTWARE\Microsoft\InetStp'
+    f_ACCESS_TYPE = Win32::Registry::KEY_READ | 0x100
```
Iristyle: IIS 7.5 needs 2008R2+ ... Server SKUs from that point are only 64-bit. We don't allow 32-bit agents on 64-bit anymore.

Windows desktop SKUs can still be 32-bit, but we don't have to worry about reg redirection.

So I think this access is correct.... as long as  you can't install 32-bit IIS on 64-bit Windows (making it necessary to look at `Wow6432Node`)

lib/puppet/feature/iis_compatible_version_installed.rb
```diff
@@ -0,0 +1,43 @@
...
+def read_registry_for_iis_compatible_version
+  compatible_versions = ['7.5', '8.0', '8.5', '10.0']
+  installed_version = '0.0'
+
+  Puppet.debug "Reading Registry for IIS Version"
+
+  begin
+    require 'win32/registry'
+
+    f_HKLM        = Win32::Registry::HKEY_LOCAL_MACHINE
```
Iristyle: Any reason for `f_` on all the variables?

lib/puppet/feature/iis_compatible_version_installed.rb
```diff
@@ -0,0 +1,43 @@
...
+    f_REG_PATH    = 'SOFTWARE\Microsoft\InetStp'
+    f_ACCESS_TYPE = Win32::Registry::KEY_READ | 0x100
+    f_REG_KEY     = 'VersionString'
+
+    f_HKLM.open(f_REG_PATH, f_ACCESS_TYPE) do |reg|
+      installed_version = reg[f_REG_KEY][8,3]
```
Iristyle: I don't think this works for `10.0` given it's 4 characters long.

Instead, I think you can do something like:

```ruby
installed_version = Gem::Version.new(reg[f_REG_KEY][8..-1]).segments[0, 2].join('.')
```

lib/puppet/feature/iis_compatible_version_installed.rb
tkishel: Done

lib/puppet/feature/iis_compatible_version_installed.rb
tkishel: The original PR reused code from lib/facter/iis_version.rb
I've rebased to include the changes in MODULES-5801 and MODULES-5954.

lib/puppet/feature/iis_compatible_version_installed.rb
tkishel: Addressed via changes in MODULES-5801
See also: lib/puppet/provider/templates/webadministration/logproperties.ps1.erb

tkishel:
Upon review, I've renamed the feature to `iis_compatible_version`

I also identified a new, similar confine in lib/puppet/provider/iis_site/iisadministration.rb

```
confine :iis_version => ['10']
```

I don't think that provider is used, but if so, an `iis_version` feature would address this 'confine testing via a fact' issue in that provider. Towards that end, I've refactored iis_compatible_version.rb to define a `iis_installed_version` method that is used by the `iis_compatible_version` feature and could also be used by an `iis_version` feature.

tkishel:
Implemented in https://github.com/puppetlabs/puppetlabs-iis/pull/151

PR #108, opened 2017-09-04
(MODULES-5576) Fix: iis_site Preload Enabled Not Working

ferventcoder:
Previously, the preload enabled setting for IIS Site was setting `serviceAutostartEnabled`
instead of preload enabled. Ensure that the setting sets 'applicationDefaults.preloadEnabled'
instead.

### DIFF
--- a/lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb
@@ -22,5 +22,6 @@ Get-WebSite | %{
     logtruncatesize      = [string]$_.LogFile.truncateSize
     loglocaltimerollover = [string]$_.LogFile.localTimeRollover
     logextfileflags      = [string]$_.LogFile.logExtFileFlags
+    preloadenabled       = [string](Get-ItemProperty -Path "IIS:\Sites\$($_.Name)" -Name 'applicationDefaults.preloadEnabled' -ErrorAction 'Continue').Value
   }
 } | ConvertTo-Json -Depth 10
--- a/lib/puppet/provider/templates/webadministration/generalproperties.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/generalproperties.ps1.erb
@@ -61,7 +61,7 @@ Try-SetItemProperty @logParams
 <% if !resource[:preloadenabled].nil? -%>
 $logParams = @{
   Path  = "IIS:\Sites\$($resource.name)"
-  Name  = 'applicationDefaults.serviceAutoStartEnabled'
+  Name  = 'applicationDefaults.preloadEnabled'
   Value = '<%= "#{resource[:preloadenabled]}" %>'
 }
 Try-SetItemProperty @logParams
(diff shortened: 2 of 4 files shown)


jpogran:
You need to change the PR title one more time for this to be merged in ;)

jpogran:
I'll update the ticket to add an acceptance test for this. If you get to it before I do, I appreciate it

ferventcoder:
I think this is set so that you can add more to the PR. Go for it on the acceptance test. I assumed since it didn't fail on the change... ;)

ferventcoder:
yay, tests!

jpogran:
this is ready for review

lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb
```diff
@@ -22,5 +22,6 @@ Get-WebSite | %{
     logtruncatesize      = [string]$_.LogFile.truncateSize
     loglocaltimerollover = [string]$_.LogFile.localTimeRollover
     logextfileflags      = [string]$_.LogFile.logExtFileFlags
+    preloadenabled       = [string](Get-ItemProperty -Path "IIS:\Sites\$($_.Name)" -Name 'applicationDefaults.preloadEnabled' -ErrorAction 'Continue').Value
```
glennsarti, line 25: So this looks like the preloadEnabled element should really be defined on the application not at the site level.  Am I missing something here?

lib/puppet/provider/templates/webadministration/generalproperties.ps1.erb
```diff
@@ -61,7 +61,7 @@ Try-SetItemProperty @logParams
 <% if !resource[:preloadenabled].nil? -%>
 $logParams = @{
   Path  = "IIS:\Sites\$($resource.name)"
-  Name  = 'applicationDefaults.serviceAutoStartEnabled'
+  Name  = 'applicationDefaults.preloadEnabled'
```
glennsarti, line 64: 👍

lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb
glennsarti: Also this is setting the default of all applications within this site, whereas shouldn't we be modifying the specific application?

glennsarti: Although I can see this gives us parity with the DSC implementation.

glennsarti (commented):
I think this is ok if we're looking for DSC parity

lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb
jpogran: There are two different settings, application and site. I'm not understanding the comment here

glennsarti: I guess my point was.  This is setting the application defaults, but there's corresponding parameter on the iis_application type for preloadenabled.  I guess this is ok so at least the bug is fixed as opposed to a new parameter too.

glennsarti (approved):

ferventcoder:
I'd click merge but am unable to now ;)

PR #111, opened 2017-09-13
(Modules-5576) Fix IIS 7.5 Support

jpogran:
This fixes the iis_site provider by guarding against checking if preloadenabled has a value on IIS 7.5


lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -153,7 +153,7 @@ def self.instances
       site_hash[:loglocaltimerollover] = to_bool(site['loglocaltimerollover'])
       site_hash[:logformat]            = site['logformat']
       site_hash[:logflags]             = site['logextfileflags'].split(/,\s*/).sort
-      site_hash[:preloadenabled]       = to_bool(site['preloadenabled'])
+      site_hash[:preloadenabled]       = to_bool(site['preloadenabled']) if !site['preloadenabled'].nil?
```
glennsarti: Typical ruby usage would be `unless`

``` ruby
site_hash[:preloadenabled]       = to_bool(site['preloadenabled']) unless site['preloadenabled'].nil?
```

PR #100, opened 2017-08-09
(MODULES-5443) iis name validation

tkishel:
With this commit, iis_site :name accepts periods, which are valid characters.
Moved the logic to a common method, as iis_site :name is used by other resources.

PR #113, opened 2017-09-15
(MODULES-5443) Fix IIS Site Name Validation

michaeltlombardi:
This commit updates the name property of the iis_site type to allow
for the use of a period ('.') in the name. It also adjusts the spec
to verify this.


Iristyle:
This appears to have superseded #100

PR #119, opened 2017-10-25
(MODULES-5626) IIS Application Pool Name Validation

jpogran:
This commit fixes the name validation for iis_application_pool by being
less restrictive in the regex for name, and allowing characters like
spaces and periods.


michaeltlombardi:
Looks good to me!

michaeltlombardi (approved):

PR #120, opened 2017-10-25
(MODULES-5627) Fix IIS application name validation

jpogran:
This commit fixes the validation for applicationname and sitename by
being less restrictive to what is accepted, namely periods and spaces


michaeltlombardi (approved):
Looks good, LGTM!

PR #123, opened 2017-10-27
(MODULES-5801)(MODULES-5803) Add IIS 10.0 support

glennsarti:
Previously the IIS version detection would report that IIS was version `10.` on
Server 2016 and Windows 10.  This commit changes the IIS version detection to
use a simple regex.

This commit adds IIS Version 10.0 to the allowed confines list for the following
providers using the Web Administration model:
- iis_application
- iis_application_pool
- iis_site
- iis_virtual_directory

Previously the Preload Enabled feature would fail to be set or retrieved because
the text based version check would think that IIS version 10.0 was older than
IIS 7.5.  This commit changes the version checking to use floating point numbers
(Double) instead of strings which will then compute correctly that 10.0 is
younger than 7.5.

Previously the acceptance tests used the Web-Mgmt-Console feature for a test
fixture, however this is do not exist on Server Core.  This commit changes the
test fixture feature to Web-Scripting-Tools which is available in Core and
Full Desktop experience editions.


jpogran (approved):

clairecadman (commented):
Docs review complete

PR #132, opened 2017-11-20
(MODULES-5264) allow UNC PhysicalPath in iis_virtual_directory

tkishel:
Adds the 'Path` class to allow reuse of an accurate path validation regex  since paths are attributes of multiple types. Adds methods to test whether a path is local or remote, and if a path is local, whether the path exists.

Refactors the type to use the Path class for validation. Refactors the provider to use path methods for validation, allows the provider to create and update virtual directories with UNC paths. Adds unit tests.

### DIFF
--- a/lib/puppet/provider/iis_common.rb
+++ b/lib/puppet/provider/iis_common.rb
@@ -1,0 +1,18 @@
+def is_local_path(path)
+  return (path =~ /^.:(\/|\\)/)
+end
+
+def is_unc_path(path)
+  return (path =~ /^\\\\[^\\]+\\[^\\]+/)
+end
+
+def verify_physicalpath
+  if @resource[:physicalpath].nil? or @resource[:physicalpath].empty?
+    fail("physicalpath is a required parameter")
+  end
+  if is_local_path(@resource[:physicalpath])
+    if ! File.exists?(@resource[:physicalpath])
+      fail("physicalpath doesn't exist: #{@resource[:physicalpath]}")
+    end    
+  end
+end
--- a/lib/puppet/provider/iis_virtual_directory/webadministration.rb
+++ b/lib/puppet/provider/iis_virtual_directory/webadministration.rb
@@ -1,3 +1,4 @@
+require File.join(File.dirname(__FILE__), '../../../puppet/provider/iis_common')
 require File.join(File.dirname(__FILE__), '../../../puppet/provider/iis_powershell')
 
 Puppet::Type.type(:iis_virtual_directory).provide(:webadministration, parent: Puppet::Provider::IIS_PowerShell) do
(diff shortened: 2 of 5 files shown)


lib/puppet/provider/iis_common.rb
```diff
@@ -0,0 +1,18 @@
...
+def is_unc_path(path)
+  return (path =~ /^\\\\[^\\]+\\[^\\]+/)
+end
+
+def verify_physicalpath
+  if @resource[:physicalpath].nil? or @resource[:physicalpath].empty?
+    fail("physicalpath is a required parameter")
+  end
+  if is_local_path(@resource[:physicalpath])
+    if ! File.exists?(@resource[:physicalpath])
```
Iristyle, line 14: No existence checks for UNC paths 👍

tkishel:
This PR follows the comments in PE-22667.

lib/puppet/provider/iis_virtual_directory/webadministration.rb
```diff
@@ -40,25 +41,32 @@ def create
...
-      fail "physicalpath doesn't exist: #{@resource[:physicalpath]}"
-    end
+    verify_physicalpath
 
     cmd = []
 
-    cmd << "Set-ItemProperty -Path 'IIS:\\Sites\\#{@resource[:sitename]}\\#{@resource[:name]}' -Name 'site' -Value '#{@resource[:sitename]}';" if @resource[:sitename]
-    cmd << "Set-ItemProperty -Path 'IIS:\\Sites\\#{@resource[:sitename]}\\#{@resource[:name]}' -Name 'physicalpath' -Value '#{@resource[:physicalpath]}';" if @resource[:physicalpath]
-    cmd << "Set-ItemProperty -Path 'IIS:\\Sites\\#{@resource[:sitename]}\\#{@resource[:name]}' -Name 'application' -Value '#{@resource[:application]}';" if @resource[:application]
+    # Set-ItemProperty fails when setting site or application:
```
jpogran: Does this mean the name of the site or the application name cannot be changed after creation? If so, we need change them from parameters to properties

lib/puppet/provider/iis_virtual_directory/webadministration.rb
```diff
@@ -40,25 +41,32 @@ def create
   end
 
   def update
+    if (@resource[:ensure] == :absent)
```
jpogran: This still seems like it should be a separate ticket/fix as it doesn't haven anything to do with path validation, its fixing a core part of the provider that shouldn't be firing. I think we can let this stay in so long a new bug ticket is added for this to investigate why this provider is acting like you report.

lib/puppet_x/puppetlabs/iis/property/path.rb
```diff
@@ -0,0 +1,17 @@
+# (c:\directory or c:/directory) or \\server\share
```
jpogran: Move the comment closer to the regex, or remove from top of file

spec/unit/puppet/type/iis_virtual_directory_spec.rb
```diff
@@ -49,24 +34,75 @@
     expect(params + [:provider]).to include(*type_class.parameters)
   end
 
-  it 'should default ensure to present' do
-    pool = type_class.new(
-      name: 'foo',
-    )
-    expect(pool[:ensure]).to eq(:present)
+  let(:resource) { described_class.new(:name => "iis_virtual_directory") }
```
jpogran, line 37: Great job cleaning up these tests

jpogran (changes requested):
Minimal work needed. Great job otherwise

lib/puppet/provider/iis_virtual_directory/webadministration.rb
tkishel: MODULES-6062 documents the update after destroy issue. In response to the request in an earlier version of this PR, I added `ErrorAction Stop` to these PowerShell commands. They will throw an error if called on a deleted virtual directory.

lib/puppet/provider/iis_virtual_directory/webadministration.rb
tkishel: Yes, site and application cannot be modified with Set-ItemProperty. MODULES-6061 documents this issue. I decided to limit the scope of this PR by commenting out those commands so they do not interfere with the command that can modify physicalpath.

lib/puppet_x/puppetlabs/iis/property/path.rb
tkishel: Deleted.

jpogran (approved):

jpogran:
Adhoc passed

PR #133, opened 2017-11-21
(MODULES-6069) iis_site applicationpool does not allow valid characters

tkishel:
The fix for applicationpool was omitted from the PRs that replaced PR 100.


jpogran (commented):
Approved, pending clean adhoc run

jpogran:
Adhoc passed

PR #134, opened 2017-11-27
(MODULES-6195) Add user_name and password support to iis_virtual_directory

tkishel:
This commit adds user_name and password properties to iis_virtual_directory.
These are optional, and needed when IIS needs to impersonate an identity to
access the physical path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,6 +778,14 @@ The physical path to the virtual directory. This path must be fully qualified.
 
 The application under which the virtual directory is created.
 
+##### `user_name`
+
+The identity that should be impersonated when accessing the physical path. Optional.
+
+##### `password`
+
+The password associated with the `user_name` property. Optional.
+
 
 ## Limitations
 
--- a/lib/puppet/type/iis_virtual_directory.rb
+++ b/lib/puppet/type/iis_virtual_directory.rb
@@ -1,5 +1,6 @@
 require 'puppet/parameter/boolean'
 require_relative '../../puppet_x/puppetlabs/iis/property/path'
+require_relative '../../puppet_x/puppetlabs/iis/property/string'
 
 Puppet::Type.newtype(:iis_virtual_directory) do
   @doc = "Manage an IIS virtual directory."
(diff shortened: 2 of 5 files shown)


lib/puppet/type/iis_virtual_directory.rb
```diff
@@ -46,6 +47,14 @@
     end
   end
 
+  newproperty(:user_name, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
+    desc "Specifies the identity that should be impersonated when accessing the physical path."
+  end
+
+  newproperty(:password, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
+    desc "Specifies the password associated with the user_name property."
```
jpogran, line 55: Should we add a step to the validate block to ensure a `password` is supplied if `user_name` is specified?

tkishel: We could ... like the following?

```
  validate do
    if (self[:user_name].nil? or self[:user_name].empty?) && self[:password]
      raise ArgumentError, "A user_name is required when specifying password."
    end
    if (self[:password].nil? or self[:password].empty?) && self[:user_name]
      raise ArgumentError, "A password is required when specifying user_name."
    end
  end
```

jpogran: Something more like line 103

tkishel: But this file has only 61 lines? I'm following this to perform validation across multiple properties:  https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource_types.md#validate

jpogran: I can't count and doing too many PR reviews at once. I'm fine with your suggestion. If you can't add it today, I'll add it so we can close out the 4.2.1 release this week

tkishel: Word. 
Added.

lib/puppet/type/iis_virtual_directory.rb
```diff
@@ -46,7 +47,25 @@
...
+  end
+
   autorequire(:iis_application) { self[:application] }
   autorequire(:iis_site) { self[:sitename] }
 
+  validate do
+    if self[:user_name].to_s.empty? && !self[:password].to_s.empty?
```
Iristyle: Could simplify as 

```ruby
validate do
  # skip additional checks if both are empty
  unless self[:user_name].to_s.empty? && self[:password].to_s.empty?
    raise ArgumentError, "A user_name is required when specifying password." if self[:user_name].to_s.empty?
    raise ArgumentError, "A password is required when specifying user_name." if self[:password].to_s.empty?
  end
end
```

README.md
```diff
@@ -778,6 +778,14 @@ The physical path to the virtual directory. This path must be fully qualified.
 
 The application under which the virtual directory is created.
 
+##### `user_name`
+
+The identity that should be impersonated when accessing the physical path. Optional.
+
+##### `password`
+
+The password password associated with the `user_name` property. Optional.
```
Iristyle: Nit `password password`

lib/puppet/type/iis_virtual_directory.rb
```diff
@@ -46,7 +47,22 @@
     end
   end
 
+  newproperty(:user_name, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
+    desc "Specifies the identity that should be impersonated when accessing the physical path."
+  end
+
+  newproperty(:password, :parent => PuppetX::PuppetLabs::IIS::Property::String) do
```
Iristyle, line 54: One thing that we're going to want to do here is use the `Sensitive` type so passwords don't show up in logs.  Sensitive was added in Puppet 4.6, so this module should support it.

Given there are other passwords on other types, I will file another ticket for doing a sweep for sensitive properties that should be using `Sensitive`.

PR #136, opened 2017-11-28
(MODULES-6062) Fix iis_virtual_directory idempotency

jpogran:
This commit fixes the iis_virtual_directory destroy method to provide
the required paramters to the Remove-WebVirtualDirectory command so that
it successfully removes a virtual directory.

It also uses `@property_hash` instead of `@resource` inside the destroy
method, so that the flush method works properly after removing a virtual
directory.

Lastly, it updates the spec acceptance test to properly create and
destroy virtual directories instead of failing silently.

PR #137, opened 2017-11-29
(maint) Remove iis_virtual_directory destroy check

Iristyle:
- The PowerShell helper at _getvirtualdirectories.ps1.erb always
   returns an application of '/' when one is not specified.

   When the .instances method enumerates existing virtual directories
   it creates new provider instances, filling @property_hash with
   the values returned from the PowerShell helper.

   Therefore @property_hash[:application] should never be nil and the
   check here is not necessary.


Iristyle:
I checked this out a few ways locally while working on #136 - would be useful for someone else to try and break it.

jpogran (approved):

PR #138, opened 2017-11-29
(MODULES-6166) virtual_directory requires sitename

Iristyle:
- Previously `puppet resource iis_virtualdirectory ensure=present`
   would report success even though it failed, when `sitename` was
   left unspecified:

```
bundle exec puppet resource iis_virtual_directory foo physicalpath=c:/foo ensure=present --modulepath .\spec\fixtures\modules\
Notice: /Iis_virtual_directory[foo]/ensure: created
iis_virtual_directory { 'foo':
  ensure => 'absent',
}
```

Asking for a list of iis_virtual_directory would yield no results

```
bundle exec puppet resource iis_virtual_directory --modulepath .\spec\fixtures\modules\
```

 - This change introduces a hard failure when sitename is not specified

```
bundle exec puppet resource iis_virtual_directory foo physicalpath=c:/foo ensure=present --modulepath .\spec\fixtures\modules\
Error: sitename is a required parameter
Error: /Iis_virtual_directory[foo]/ensure: change from 'absent' to 'present' failed: sitename is a required parameter
iis_virtual_directory { 'foo':
  ensure => 'absent',
}
```

### DIFF
--- a/lib/puppet/provider/iis_virtual_directory/webadministration.rb
+++ b/lib/puppet/provider/iis_virtual_directory/webadministration.rb
@@ -24,7 +24,8 @@ Puppet::Type.type(:iis_virtual_directory).provide(:webadministration, parent: Pu
     cmd = []
     if is_local_path(@resource[:physicalpath])
       cmd << "New-WebVirtualDirectory -Name \"#{@resource[:name]}\" "
-      cmd << "-Site \"#{@resource[:sitename]}\" " if @resource[:sitename]
+      fail("sitename is a required parameter") unless @resource[:sitename]
+      cmd << "-Site \"#{@resource[:sitename]}\" "
     else
       # New-WebVirtualDirectory fails when PhysicalPath is a UNC path that unavailable,
       # and UNC paths are inherently not necessarily always available.


Iristyle:
I didn't do a ton of validation on this, but wanted to throw up at least a WIP PR for the problem that I'm aware of.

Iristyle:
~~Discovered that the same check needs to be performed in `destroy` or bad things can potentially happen.  In the case where sitename is omitted, the virtual directory name is put into the path in such a way that it takes the place of the sitename - meaning there is potential to do harm to the site itself.~~

UPDATE: comment above was prior to #136 being merged and is no longer valid

jpogran:
If the type requires a non empty sitename, why does puppet resource allow an empty sitename?

Iristyle:
@jpogran `puppet resource` won't call the per-property `validate` blocks defined in the type. You can add a global `validate` for the entire type, but then there's another interesting bug that you will run into - documented in https://tickets.puppetlabs.com/browse/PUP-3732

Iristyle:
Discovered that thanks to https://github.com/puppetlabs/puppetlabs-iis/commit/00096d2541bbab4f41777b9b447d919203c84aa2#diff-c5a995fbaf1a8c6f0d0bf229efc84fbaR63, adding a check in `destroy` is not necessary anymore.

This PR should be sufficient to handle this case now @jpogran

Iristyle:
Another way to plausibly address is this is to make sure that any failures to execute the PowerShell commands return a failure status during `create` so that Puppet understands the resource wasn't created.

If the PowerShell code being run set `$ErrorActionPreference = 'Stop'` to immediately bail and return what could be interpreted as a non-zero exit code, then additional validation here *probably* wouldn't be necessary (needs to be tested for sure)

RandomNoun7 (approved):

PR #140, opened 2017-12-01
(MODULES-6150) Prepare 4.2.1 release redux

Iristyle:


README.md
```diff
@@ -772,7 +772,7 @@ The site under which the virtual directory is created.
 
 ##### `physicalpath`
 
-The physical path to the virtual directory. This path must be fully qualified.
+The physical path to the virtual directory. This path must be fully qualified. Though not recommended, this may be a UNC style path. Credentials for access to the UNC path may be supplied by the `user_name` and `password` properties.
```
clairecadman: I would suggest changing this to:

Though not recommended, this can be a UNC style path. Supply credentials for access to the UNC path with the `user_name` and `password` properties.

Iristyle: Active vs passive, of course... thanks!

ThoughtCrhyme (approved):

PR #142, opened 2017-12-11
Update README.md

clairecadman:

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ iis_virtual_directory { 'vdir':
 
 Allows creation of a new IIS Application and configuration of application parameters.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -163,6 +163,8 @@ iis_virtual_directory { 'vdir':
 
 Allows creation of a new IIS Application and configuration of application parameters.
 
+The iis_application type uses a composite namevar for applicationname and sitename to uniquely identify a declaration. To use this successfully, put both the sitename and the applicationname in the title. Puppet will build the catalog using the composite of the two values, while still using the correct value for the applicationname when creating the IIS application.
```
jpogran: It may be useful to put an example here, as this doesn't say it requires a `/` in between the `sitename` and `applicationname`

jpogran (changes requested):

clairecadman:
@jpogran Please see update.

jpogran (approved):

PR #150, opened 2018-01-11
(MODULES-6385) Create site with alternate port

RandomNoun7:
This change will ensure that a user can avoid collisions on port 80 when
creating a new website. Prior to this change when the module created a new 
website, it was not able to see the port binding information when it called the
New-Website cmdlet. This meant that if a website already existed with a binding
to port 80, the site would fail to start. Since New-Website always attempts to
start the website after it is created, this would cause the whole manifest to
fail. With this change, if a binding other than port 80 is specified, that
binding is used in the call to New-Website, and the conflict is avoided.

### DIFF
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -168,4 +168,11 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
     return :false  if value == false  || value =~ (/(^$|false|f|no|n|0)$/i)
     raise ArgumentError.new("invalid value for Boolean: \"#{value}\"")
   end
+
+  def binding_port
+    if @resource[:bindings]
+      binding = @resource[:bindings].first
+      return binding['bindinginformation'].match(/:([0-9]*):/).captures.first
+    end
+  end
 end
--- a/spec/acceptance/iis_site_spec.rb
+++ b/spec/acceptance/iis_site_spec.rb
@@ -545,5 +545,53 @@ describe 'iis_site' do
         remove_all_sites
       end
     end
+
+    context 'with conflicting sites' do
+      before (:all) do
+        create_path('C:\inetpub\basic')
+        @site_name = "#{SecureRandom.hex(10)}"
+        @second_site_name = "#{SecureRandom.hex(10)}"
+        create_site(@site_name, true)
+
+        @manifest = <<-HERE
+          iis_site { "#{@second_site_name}":
+            ensure          => 'started',
+            physicalpath    => 'C:\\inetpub\\basic',
+            applicationpool => 'DefaultAppPool',
+            bindings        => [
+              {
+                'bindinginformation' => "*:8080:#{@second_site_name}",
+                'protocol'           => 'http',
+              }
+            ],
+          }
+        HERE
+      end
+
+      it_behaves_like 'an idempotent resource'
+
+      context 'when puppet resource is run' do
+        before(:all) do
+          @first_site = resource('iis_site', @site_name)
+          @second_site = resource('iis_site', @second_site_name)
+        end
+
+        it "should run the first site on port 80" do
+          expect(@first_site.stdout).to match(/ensure(\s*)=> 'started',/)
+
+          # binding info for this site does not contain a host name.
+          expect(@first_site.stdout).to match(/\*\:80\:/)
+        end
+
+        it "should run the second site on port 8080" do
+          expect(@second_site.stdout).to match(/ensure(\s*)=> 'started',/)
+          expect(@second_site.stdout).to match(/\*\:8080\:#{@second_site_name}/)
+        end
+      end
+
+      after(:all) do
+        remove_all_sites
+      end
+    end
   end
 end
--- a/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
@@ -11,6 +11,7 @@ $createParams = @{
   ApplicationPool = $resource.applicationpool
   Force           = $true
   ErrorAction     = 'Stop'
+<%= "port            = #{resource.provider.binding_port}" unless resource.provider.binding_port.nil? %>
 }
 
 # If there are no other websites, specify the Id, otherwise an Index Out of Range error can be thrown


RandomNoun7:
TODO: Additional tests to cover the change.

spec/acceptance/iis_site_spec.rb
```diff
@@ -539,11 +539,41 @@
         HERE
       end
 
+      require 'pry'; binding.pry;
```
michaeltlombardi: This is still WIP, but need to pull the pry before merge.

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -168,4 +168,11 @@ def self.to_bool(value)
     return :false  if value == false  || value =~ (/(^$|false|f|no|n|0)$/i)
     raise ArgumentError.new("invalid value for Boolean: \"#{value}\"")
   end
+
+  def binding_value
```
Iristyle: A name that included `port` would be helpful I think, given you're fishing the port out of the binding string?

spec/acceptance/iis_site_spec.rb
```diff
@@ -545,5 +545,34 @@
...
+              {
+                'bindinginformation' => "*:8080:#{@second_site_name}",
+                'protocol'           => 'http',
+              }
+            ],
+          }
+        HERE
+      end
+
+      it_behaves_like 'an idempotent resource'
```
Iristyle, line 571: So I looked at the other tests and saw some similar ones already that specified `bindinginformation` like this that didn't fail.

If you go back to the top of the test file, you'll see that entire suite starts off with a `remove_all_sites` to remove the Default binding:

https://github.com/RandomNoun7/puppetlabs-iis/blob/57fd6505f2a9d70a68fcfc97abd846329439cc99/spec/acceptance/iis_site_spec.rb#L6

So you have a couple of options:

* Put this test *before* that happens to prove this works. You would also want to add an assertion to this test that confirming another site exists on 80. I'd recommend against this as it relies on global state that could change in the test file, introduces a weird ordering, etc.
* Modify the setup of this test to add a different site on 80, and assert that it's there (its state of running / stopped shouldn't matter).  Then you can verify no errors get thrown trying to create a new site because you're creating a new site bound to a different port instead of default binding it to 80 first.

Iristyle (changes requested):

spec/acceptance/iis_site_spec.rb
Iristyle: To follow up on this, @RandomNoun7 noted I missed the bit about `create_site` that actually sets up a site on 80 as part of the `before(:all)`.  So it looks like there *is* a site running on 80 prior to running the manifest, which would have previously conflicted.

For the sake of understanding the test, it would be helpful to add a couple of additional assertions around the parts that are most important for this test. To the `before` we want to add:

```
@result = resource('iis_site', @site_name)
```

So that before the `it_behaves_like 'an idempotent resource'` we can add an assertion confirming `@site_name` is running on `80`:

```ruby
puppet_resource_should_show('port', 80)
```

Then after the idempotent check that runs the manifest:


```ruby
@result = resource('iis_site', @second_site_name)
puppet_resource_should_show('port', 8080)
```

I think it's preferable to be able to check the state outside of what `puppet resource` tells us, but if that's the easy way, it's already an established pattern in the test file.

RandomNoun7: New tests added.

spec/acceptance/iis_site_spec.rb
```diff
@@ -545,5 +545,53 @@
...
+        it "should run the first site on port 80" do
+          expect(@first_site.stdout).to match(/ensure(\s*)=> 'started',/)
+
+          # binding info for this site does not contain a host name.
+          expect(@first_site.stdout).to match(/\*\:80\:/)
+        end
+
+        it "should run the second site on port 8080" do
+          expect(@second_site.stdout).to match(/ensure(\s*)=> 'started',/)
+          expect(@second_site.stdout).to match(/\*\:8080\:#{@second_site_name}/)
```
RandomNoun7, line 588: Decided to remove calls to `puppet_resource_should_show()` entirely. That function relies on the hard coded `@resource` variable being populated, which means that if I want to test the properties of two resources, one would be able to use the `@resource` variable, and the other would have to use another variable with manual regex matching. I decided that was not a great way to handle this and that neither should use that method.

Iristyle: Yeah, that automatic behavior around `@resource` is an interesting design choice in the test helpers.

Iristyle:
Minor - need to update the commit message to start with `(MODULES-6385)`

Iristyle (approved):

RandomNoun7:
Commit message edited.

Iristyle (approved):
This all looks good to me now... are there any reservations you had around lifecycle management @jpogran ?

Iristyle:
Do you want to add Beaker output as a comment to the PR to verify tests ran successfully @RandomNoun7 ?

RandomNoun7:
Code output from new tests below:
```
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:19:14$ cmd.exe /c puppet apply --verbose --detailed-exitcodes C:/cygwin64/tmp/apply_manifes
.pp.aBDnWk                                                                                                                                                    
  Info: Loading facts                                                                                                                                         
  Notice: Compiled catalog for n7r82qdtt8gtkjp.delivery.puppetlabs.net in environment production in 0.27 seconds                                              
  Info: Applying configuration version '1516144763'                                                                                                           
  Notice: /Stage[main]/Main/Iis_site[d74b41d4a4b21e3b5418]/ensure: defined 'ensure' as 'started'                                                              
  Notice: Applied catalog in 10.69 seconds                                                                                                                    
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 21.38 seconds                                                                     
Exited: 2                                                                                                                                                     
        should run without errors                                                                                                                             
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:19:35$ cygpath -m $(mktemp -t apply_manifest.pp.XXXXXX)                                    
  C:/cygwin64/tmp/apply_manifest.pp.Pberi8                                                                                                                    
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 0.17 seconds                                                                      
localhost $ scp /tmp/beaker20180116-3202-1bcesmu windows2016_core-64-1:C:/cygwin64/tmp/apply_manifest.pp.Pberi8 {:ignore => }                                 
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:19:36$ cmd.exe /c puppet apply --verbose --detailed-exitcodes C:/cygwin64/tmp/apply_manifes
.pp.Pberi8                                                                                                                                                    
  Info: Loading facts                                                                                                                                         
  Notice: Compiled catalog for n7r82qdtt8gtkjp.delivery.puppetlabs.net in environment production in 0.23 seconds                                              
  Info: Applying configuration version '1516144785'                                                                                                           
  Notice: Applied catalog in 3.02 seconds                                                                                                                     
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 13.41 seconds                                                                     
        should run a second time without changes                                                                                                              
      when puppet resource is run                                                                                                                             
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:19:49$ cmd.exe /c puppet resource iis_site 916e23a4e51c52d11ac4                            
  iis_site { '916e23a4e51c52d11ac4':                                                                                                                          
    ensure               => 'started',                                                                                                                        
    applicationpool      => 'DefaultAppPool',                                                                                                                 
    bindings             => [{'bindinginformation' => '*:80:', 'protocol' => 'http'}],                                                                        
    enabledprotocols     => 'http',                                                                                                                           
    limits               => {'connectiontimeout' => '120', 'maxbandwidth' => '4294967295', 'maxconnections' => '4294967295'},                                 
    logflags             => ['ClientIP', 'Date', 'HttpStatus', 'HttpSubStatus', 'Method', 'Referer', 'ServerIP', 'ServerPort', 'Time', 'TimeTaken', 'UriQuery'
 'UriStem', 'UserAgent', 'UserName', 'Win32Status'],                                                                                                          
    logformat            => 'W3C',                                                                                                                            
    loglocaltimerollover => 'false',                                                                                                                          
    logpath              => '%SystemDrive%\inetpub\logs\LogFiles',                                                                                            
    logperiod            => 'Daily',                                                                                                                          
    logtruncatesize      => '20971520',                                                                                                                       
    physicalpath         => 'C:\inetpub\wwwroot',                                                                                                             
    preloadenabled       => 'false',                                                                                                                          
  }                                                                                                                                                           
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 11.56 seconds                                                                     
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:20:01$ cmd.exe /c puppet resource iis_site d74b41d4a4b21e3b5418                            
  iis_site { 'd74b41d4a4b21e3b5418':                                                                                                                          
    ensure               => 'started',                                                                                                                        
    applicationpool      => 'DefaultAppPool',                                                                                                                 
    bindings             => [{'bindinginformation' => '*:8080:d74b41d4a4b21e3b5418', 'protocol' => 'http'}],                                                  
    enabledprotocols     => 'http',                                                                                                                           
    limits               => {'connectiontimeout' => '120', 'maxbandwidth' => '4294967295', 'maxconnections' => '4294967295'},                                 
    logflags             => ['ClientIP', 'Date', 'HttpStatus', 'HttpSubStatus', 'Method', 'Referer', 'ServerIP', 'ServerPort', 'Time', 'TimeTaken', 'UriQuery'
 'UriStem', 'UserAgent', 'UserName', 'Win32Status'],                                                                                                          
    logformat            => 'W3C',                                                                                                                            
    loglocaltimerollover => 'false',                                                                                                                          
    logpath              => '%SystemDrive%\inetpub\logs\LogFiles',                                                                                            
    logperiod            => 'Daily',                                                                                                                          
    logtruncatesize      => '20971520',                                                                                                                       
    physicalpath         => 'C:\inetpub\basic',                                                                                                               
    preloadenabled       => 'false',                                                                                                                          
  }                                                                                                                                                           
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 11.12 seconds                                                                     
        should run the first site on port 80                                                                                                                  
        should run the second site on port 8080                                                                                                               
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) 23:20:12$ C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -noninteractive -noprofile
-executionpolicy bypass -command "& {Import-Module WebAdministration -ErrorAction Stop;cd iis: ;Get-Website | Remove-Website}" < /dev/null                    
                                                                                                                                                              
n7r82qdtt8gtkjp.delivery.puppetlabs.net (windows2016_core-64-1) executed in 0.59 seconds                                                                      
ssh connection to windows2016_core-64-1 has been terminated                                                                                                   
Handing 'n7r82qdtt8gtkjp.delivery.puppetlabs.net' back to vmpooler for VM destruction                                                                         
Spent 0.01 seconds cleaning up                                                                                                                                
                                                                                                                                                              
Finished in 2 minutes 7.2 seconds (files took 1 minute 28.52 seconds to load)                                                                                 
4 examples, 0 failures                                                                                                                                       
```

jpogran (approved):

PR #151, opened 2018-01-12
(MODULES-5465) add puppet feature suitability

jpogran:
(MODULES-5465) Add puppet feature suitability

This commit changes the confine for all providers to use a puppet
feature instead of a puppet fact. This allows the provider to install
IIS and then use one of the providers to configure it in the same Puppet
run. A feature is used instead of a fact because a feature can be
re-evaluated during the Puppet run and a fact cannot.

This commit also extracts the IIS detection logic out to a class in
order to isolate the parts that decide what IIS versions we support and
how to determine existance of IIS on the target node. Appropriate spec
tests are added for the class.

Automated acceptance tests were not added for this, as the
spec_helper_acceptance installs IIS before any tests are run. Since this
change is to enable installing and configuring IIS in the same run, we
would have to re-write all of the tests in order to be able to add one
test for this. This has been manually validated instead.

### DIFF
--- a/lib/facter/iis_version.rb
+++ b/lib/facter/iis_version.rb
@@ -1,24 +1,7 @@
 Facter.add("iis_version") do
  confine :kernel => :windows
   setcode do
-    iis_ver = nil
-    begin
-      require 'win32/registry'
-      access_type = Win32::Registry::KEY_READ | 0x100
-      hklm        = Win32::Registry::HKEY_LOCAL_MACHINE
-      reg_path    = 'SOFTWARE\Microsoft\InetStp'
-      reg_key     = 'VersionString'
-      
-      iis_version_text = ''
-      hklm.open(reg_path, access_type) do |reg|
-        iis_version_text = reg[reg_key]
-      end
-      if iis_version_text.match(/^Version (\d+\.\d+)$/)
-        iis_ver = $1
-      end
-    rescue
-      iis_ver = nil
-    end
-    iis_ver
+    require_relative '../puppet_x/puppetlabs/iis/iis_version'
+    PuppetX::PuppetLabs::IIS::IISVersion::installed_version
   end
 end
--- a/spec/unit/puppet_x/puppetlabs/iis/iis_version_spec.rb
+++ b/spec/unit/puppet_x/puppetlabs/iis/iis_version_spec.rb
@@ -1,0 +1,62 @@
+require 'spec_helper'
+require 'puppet/type'
+require 'puppet_x/puppetlabs/iis/iis_version'
+
+describe PuppetX::PuppetLabs::IIS::IISVersion, :if => Puppet::Util::Platform.windows? do
+  before(:each) do
+    @ps = PuppetX::PuppetLabs::IIS::IISVersion
+  end
+
+  describe "when iis is installed" do
+
+    it "should detect a iis version" do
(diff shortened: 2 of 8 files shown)


jpogran:
Question: how do we add an automated test to test suitability when the test harness installs the thing we need to install in the test, and uninstalling requires a reboot of the host?

lib/facter/iis_version.rb
```diff
@@ -1,24 +1,7 @@
...
-        iis_version_text = reg[reg_key]
-      end
-      if iis_version_text.match(/^Version (\d+\.\d+)$/)
-        iis_ver = $1
-      end
-    rescue
-      iis_ver = nil
-    end
-    iis_ver
+    # require 'puppet_x/puppetlabs/iis/iis_version'
```
Iristyle: I think your intent here is to run the helper code to return the fact value in the future, right? Wasn't sure as the lines are commented... 

In case not, fact removal is generally considered a breaking change, at least in Puppet... so in Puppet itself we almost never remove facts, only add them... and change format / type of return value very carefully.

spec/unit/puppet_x/puppetlabs/iis/iis_version_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+describe PuppetX::PuppetLabs::IIS::IISVersion, :if => Puppet::Util::Platform.windows? do
+  before(:each) do
+    @ps = PuppetX::PuppetLabs::IIS::IISVersion
+  end
+
+  describe "when iis is installed" do
+
+    it "should detect a iis version" do
+      Win32::Registry.any_instance
```
Iristyle: Avoid `any_instance` if possible.

http://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/AnyInstance

Iristyle:
As for the suitability question... its tested implicitly by *anything* being able to run, right?

If you're interested in the negative case, easiest thing might be to guard the test and only run on a non-Windows host?

We have a similar test in DSC for Linux - https://github.com/puppetlabs/puppetlabs-dsc/blob/3773d00b7f0cd475ee848bcaee0728d7bcfb03c1/tests/integration/tests/basic_functionality/negative/dsc_on_linux.rb   .... this could probably be rewritten as an rspec test that runs in Travis. I can help out with that if you want to go that route.

lib/facter/iis_version.rb
jpogran: The former, not the latter.

jpogran:
It's tested implicitly by anything being able to run in the *same* puppet run that installs IIS. IIS is installed by the spec_acceptance_helper before the actual tests run, so any puppet run is suitable in the tests.

The only options I can think of is to move the IIS install part to each spec test, and force the first test to be suitability. Requiring a certain test order and duplicating install code does not sound like a good thing.

michaeltlombardi:
I'm okay with skipping the automated testing here to avoid introducing fragility and maintenance complexity into the test suite.

michaeltlombardi (approved):

ThoughtCrhyme:
> Requiring a certain test order and duplicating install code does not sound like a good thing.

Agreed

Iristyle:
+1 to @michaeltlombardi suggestion on punting on the test and using manual validation

PR #152, opened 2018-01-17
(MODULES-6257) Fix Applications in Nested Subfolders

RandomNoun7:
(MODULES-6257) This commit allows the user to create and manage IIS
Applications in nested subfolders. Prior to this change an application
in a nested subfolder had its name handled improperly, such that the
iis_application resource was unable to detect that the application
specified in the manifest already existed.

### DIFF
--- a/spec/acceptance/iis_application_spec.rb
+++ b/spec/acceptance/iis_application_spec.rb
@@ -80,6 +80,132 @@ describe 'iis_application' do
         remove_all_sites
       end
     end
+
+    context 'with nested virtual directory' do
+      before(:all) do
+        @site_name  = SecureRandom.hex(10)
+        @app_name   = SecureRandom.hex(10)
+        create_site(@site_name, true)
+        create_path("c:\\inetpub\\wwwroot\\subFolder\\#{@app_name}")
+        @manifest = <<-HERE
+          iis_application{'subFolder/#{@app_name}':
+            ensure => 'present',
+            applicationname => 'subFolder/#{@app_name}',
+            physicalpath => 'c:\\inetpub\\wwwroot\\subFolder\\#{@app_name}',
+            sitename => '#{@site_name}'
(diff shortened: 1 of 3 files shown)


RandomNoun7:
So those new tests are pretty ugly, but that structure was the best way I could find to avoid the scoping and order of execution issues I ran into in pulling back and testing the proper resources.

spec/acceptance/iis_application_spec.rb
```diff
@@ -80,6 +80,72 @@
...
+            applicationname => 'subFolder/#{@app_name}',
+            physicalpath => 'c:\\inetpub\\wwwroot\\subFolder\\#{@app_name}',
+            sitename => '#{@site_name}'
+          }
+        HERE
+      end
+
+      it_behaves_like 'an idempotent resource'
+
+      describe "application validation" do
```
RandomNoun7, line 102: If I don't put a ```describe``` block here, then the ```it_behaves_like``` will actually fire after the ```it``` below here. Since the ```it_behaves_like``` is the function that actually runs the manifest, that meant all of the tests in the ```it``` failed. Putting the tests in a ```describe``` was the only way I found to ensure the tests executed in the proper order.

RandomNoun7 (commented):
Just adding an explanatory comment.

ThoughtCrhyme (commented):
Using "on" instead of "resource_on" is the better approach and more consistent with other tests.

I dig it.

RandomNoun7:
Added parameter validation to the ```applicationname``` property. Added additional tests. When this version passes tests it's ready for merge.

jpogran (changes requested):
Commit message header does not have ticket number in it. Otherwise 👍

RandomNoun7:
@jpogran Done.

jpogran (approved):

PR #157, opened 2018-03-13
(MODULES-6673) Ensure Present IIS Site Loop

RandomNoun7:
This change ensures that if a user sets the Ensure property of a
manifest to present that Puppet run will neither attempt to start the
site if it is stopped nor display a state transition message to the
terminal and the logs specifying a change from <state> to present.

### DIFF
--- a/lib/puppet/type/iis_site.rb
+++ b/lib/puppet/type/iis_site.rb
@@ -20,11 +20,17 @@ Puppet::Type.newtype(:iis_site) do
     end
 
     newvalue(:present) do
-      provider.create
+      provider.create unless provider.exists?
     end
 
     newvalue(:absent) do
       provider.destroy
+    end
+
+    def insync?(is)
+      is.to_s == should.to_s or
+        (is.to_s == 'started' and should.to_s == 'present') or
+        (is.to_s == 'stopped' and should.to_s == 'present')
     end
 
     aliasvalue(:false, :stopped)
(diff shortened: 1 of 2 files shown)


RandomNoun7:
The test added to the commit ensures that you can set a sites to stopped and apply a manifest with Ensure value of Present, and it should act idempotently without messages about state changes, and it should not restart the web site.

jpogran (approved):

lib/puppet/type/iis_site.rb
```diff
@@ -20,13 +20,19 @@
...
-      provider.create
+      provider.create unless provider.exists?
     end
 
     newvalue(:absent) do
       provider.destroy
     end
 
+    def insync?(is)
+      is.to_s == should.to_s or
```
Iristyle, line 31: Prefer to use `||` and `&&` for logical or and operations, not their control-flow counterparts. Operator precedence is different.

http://www.virtuouscode.com/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/

PS: I would say the same thing about the original AWS module, which should also be changed.

Iristyle:
@RandomNoun7 what was the result of the `puppet resource` testing? Were you able to see if that behavior was already a bug or introduced by this PR? New ticket filed?

RandomNoun7:
@Iristyle The bug already existed and a ticket filed.

PR #158, opened 2018-03-13
(MODULES-6637) Port Conflict Hostname Only

RandomNoun7:
This change makes it possible to pass both the port number and the
hostname header value from the bindinginformation property to the
New-Website cmdlet when creating a new website.

### DIFF
--- a/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
+++ b/lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
@@ -1,3 +1,5 @@
+<% ip_dns, port, host_header = resource.provider.binding_information %>
+
 $resource = @{
   name            = '<%= "#{resource[:name]}" %>'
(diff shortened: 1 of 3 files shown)


michaeltlombardi:
Minor typo in commit, `Thi` should be `This`.

lib/puppet/provider/templates/webadministration/_newwebsite.ps1.erb
```diff
@@ -1,3 +1,5 @@
+<%ip_dns, port, host_header = resource.provider.binding_information%>
```
michaeltlombardi: Nit: Whitespace by the tags is preferable, I think - `<% ip_dns` etc.

michaeltlombardi (dismissed):
Only the most minor of nits. :)

RandomNoun7:
Nits squished.

michaeltlombardi (approved):

PR #160, opened 2018-03-16
(MODULES-5459) Update Documentation

RandomNoun7:
This change brings the documentation in the type files into sync with
what was present in the README.md file. In some cases the text was
better in the type file and so the two sources were combined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,15 +252,15 @@ Allows creation of a new IIS Application Pool and configuration of application p
 
 #### `ensure`
 
-Must be either 'present' or 'absent'. Present will ensure the application pool is created.
+Specifies whether an application pool should be present or absent. If `state` is not specified, the application pool will be created and left in the default started state.
 
 #### `name`
 
-Name of the application pool.
+The unique name of the ApplicationPool. Must be unique.
 
 #### `state`
 
-The state of the application pool. Must be either 'started' or 'stopped'.
+The state of the ApplicationPool. By default, a newly created application pool will be started. Valid options 'started' or 'stopped'.
 
 #### `auto_start`
 
@@ -308,7 +308,7 @@ Configures the action that IIS takes when a worker process exceeds its configure
 
 #### `cpu_limit`
 
-Configures the maximum percentage of CPU time (in 1/1000ths of one percent) that the worker processes in an application pool are allowed to consume over a period of time as indicated by the `cpu_reset_interval` property. If the limit set by the limit property is exceeded, an event is written to the event log and an optional set of events can be triggered. These optional events are determined by the `cpu_action` property.
+Configures the maximum percentage of CPU time (in 1/1000ths of one percent) that the worker processes in an application pool are allowed to consume over a period of time as indicated by the `cpu_reset_interval` property. If the limit set by the limit property is exceeded, an event is written to the event log and an optional set of events can be triggered. These optional events are determined by the `cpu_action` property. Value must be <= 100000
 
 #### `cpu_reset_interval`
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -252,15 +252,15 @@ Allows creation of a new IIS Application Pool and configuration of application p
 
 #### `ensure`
 
-Must be either 'present' or 'absent'. Present will ensure the application pool is created.
+Specifies whether an application pool should be present or absent. If `state` is not specified, the application pool will be created and left in the default started state.
 
 #### `name`
 
-Name of the application pool.
+The unique name of the ApplicationPool.
```
michaeltlombardi: Possibly rephrase this to:

> The name of the ApplicationPool. **Must** be unique.

README.md
```diff
@@ -252,15 +252,15 @@ Allows creation of a new IIS Application Pool and configuration of application p
...
 
 #### `state`
 
-The state of the application pool. Must be either 'started' or 'stopped'.
+The state of the ApplicationPool. By default, a newly created application pool will be started
```
michaeltlombardi: Suggest re-including that the value must be either started or stopped in addition to the new information.

README.md
```diff
@@ -503,7 +507,7 @@ Must be either 'present' or 'absent'.
 
 ##### `name`
 
-The name of the feature.
+The unique name of the feature to manage.
```
michaeltlombardi: As above, suggest clarifying that it must be unique as opposed to uniqueness being a intrinsic quality of the property.

ie, the difference between 'the unique name' and 'the name which must be unique' is semantically important.

README.md
```diff
@@ -536,7 +540,7 @@ If 'stopped' is specified, then the site will be created and stopped.
 
 ##### `name`
 
-The name of the site.
+The Name of the IIS site. Used for uniqueness. Will set the target to this value if target is unset.
```
michaeltlombardi: Rephrase to clarify that it must be unique to keep in line with other notes.

README.md
```diff
@@ -552,7 +556,17 @@ The protocols enabled for the site. If 'https' is specified, 'http' is implied.
 
 ##### `bindings`
 
-Specifies one or more bindings (protocol, address, port, and ssl certificate) for the site.
+The protocol, address, port, and ssl certificate bindings for a web site.
+
+The bindinginformation value should be in the form of the IPv4/IPv6 address or wildcard *, then the port, then the optional hostname separated by colons:  `(ip|\*):[1-65535]:(hostname)?`
```
michaeltlombardi: Suggest placing the wildcard in backticks to fix formatting and clarify.

lib/puppet/type/iis_application.rb
```diff
@@ -27,7 +37,8 @@ def self.title_patterns
   end
 
   newparam(:applicationname, :namevar => true) do
-    desc "The name of the Application. The virtual path of an application is /<applicationname>"
+    desc "The name of the application. The virtual path of the application is
+          '/<applicationname>'."
```
michaeltlombardi, line 41: Is this change to accommodate line length?

README.md
RandomNoun7: Extremely agree.

README.md
RandomNoun7: Agree to the fullest extent of the law.

michaeltlombardi (changes requested):
A few minor changes and tweaks.

lib/puppet/type/iis_application.rb
RandomNoun7: Yes. For a while I was wrapping at only approximately  80 chars, but it got annoying to be in the business of making a hundred judgement calls about when to wrap and when not to wrap, so I switch to a programmatic 80 char wrap.

README.md
```diff
@@ -536,7 +540,7 @@ If 'stopped' is specified, then the site will be created and stopped.
 
 ##### `name`
 
-The name of the site.
+The Name of the IIS site. Must be uniquene. Will set the target to this value if target is unset.
```
michaeltlombardi: Spelling - 'uniquene' should be 'unique'

michaeltlombardi (approved):

PR #168, opened 2018-05-01
(MODULES-6869) add a test for unicode site name

ThoughtCrhyme:
Add a test for UTF-8 site_name that expects failure such that when we fix
UTF-8 characters in IIS we'll have a working acceptance test.

### DIFF
--- a/spec/acceptance/iis_unicode_site_spec.rb
+++ b/spec/acceptance/iis_unicode_site_spec.rb
@@ -1,0 +1,53 @@
+require 'spec_helper_acceptance'
+
+describe 'iis_site' do
+  before(:all) do
+    # Remove 'Default Web Site' to start from a clean slate
+    remove_all_sites();
+  end
+
+  context 'when configuring a website' do
+    context 'with required parameters and UTF-8 site name' do
+      before (:all) do
+        create_path('C:\inetpub\basic')
+        @site_name = "\u4388\u542B\u3D3C\u7F4D\uF961\u4381\u53F4\u79C0\u3AB2\u8EDE" # 䎈含㴼罍率䎁叴秀㪲軞
+        @manifest = <<-HERE
+          iis_site { '#{@site_name}':
+            ensure          => 'started',
+            physicalpath    => 'C:\\inetpub\\basic',
+            applicationpool => 'DefaultAppPool',
+          }
+        HERE
+      end
+
+      it 'should run without errors' do
+        expect_failure('Expected to fail due to MODULES-6869') do
+          execute_manifest(@manifest, :catch_failures => true)
+        end
+      end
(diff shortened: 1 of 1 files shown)


Iristyle:
I think this generally looks good @ThoughtCrhyme ... any output from the Beaker run to have a look at?

ThoughtCrhyme:
@Iristyle output from test as written when run locally is:

```
iis_site
  when configuring a website
    with required parameters and UTF-8 site name
localhost $ scp /var/folders/hj/xqmb6zjj2tjg66bp4mhrgx3h0000gq/T/beaker20180501-75007-moh8xy windows2012-64-1:C:/cygwin64/tmp/apply_manifest.pp.nkxkSF {:ignore => }
An assertion was expected to fail, and did. This is probably due to a known product bug, and is probably not a problem. Additional info: 'Expected to fail due to MODULES-6869' Failed assertion: 'Unacceptable exit code returned: 1. Acceptable code(s): 0, 2'
      should run without errors
An assertion was expected to fail, and did. This is probably due to a known product bug, and is probably not a problem. Additional info: 'Expected to fail due to MODULES-6869' Failed assertion: 'Expected IIS site was not present!.
Expected /^1$/ to match "0\n".'
      Verify that IIS site name is present
```

When I run it with an iis_site that is ASCII and I remove the expect_failure calls the output is:

```
iis_site
  when configuring a website
    with required parameters and UTF-8 site name
localhost $ scp /var/folders/hj/xqmb6zjj2tjg66bp4mhrgx3h0000gq/T/beaker20180501-74386-a78dnt windows2012-64-1:C:/cygwin64/tmp/apply_manifest.pp.AunpKA {:ignore => }
      should run without errors
      Verify that IIS site name is present
```

Iristyle (approved):

pillarsdotnet:
Why are we getting a `NoMethodError` saying that `assert_match` is an undefined method?

> WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was #<NoMethodError: undefined method `assert_match' for #<RSpec::ExampleGroups::IisSite_2::WhenConfiguringAWebsite::WithRequiredParametersAndUTF8SiteName:0x0000000008c50370>>. Instead consider providing a specific error class or message. This message can be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`. Called from C:/projects/puppetlabs-iis-9rnb2/spec/acceptance/iis_unicode_site_spec.rb:38:in `block (4 levels) in <top (required)>'.

PR #171, opened 2018-06-08
(MODULES-7173) Cannot Start IIS Site with Port 443

RandomNoun7:
This change fixese a bug where a site bound to port 443 in the binding
information hash could not be started. Before version 4.3.0 all sites
were created on Port 80, and only later in the configuration process
would they be given a port binding to port 443 and the https protocol
enabled on the site. After 4.3.0, the port binding was passed to the
New-Website cmdlet by default as a way to avoid conflicts on port 80.
Unfortunately, the https protocol was not enabled by default if the port
binding was 443. When the site was started, as it always is by default,
IIS would throw an error because a site bound to port 443 is not allowed
to start if the HTTPS protocol is not enabled.

This change detects when the HTTPS protocol has been specified as
enabled, or when the user has specified port 443 as the port binding. In
reality checking for both is probably not needed as they are mutual
requirements, but it costs little to do the check, so both are checked.

This change also incorporates additional automated testing to ensure
that future releases do not break HTTPS sites.

### DIFF
--- a/lib/puppet/provider/iis_site/webadministration.rb
+++ b/lib/puppet/provider/iis_site/webadministration.rb
@@ -191,4 +191,11 @@ Puppet::Type.type(:iis_site).provide(:webadministration, parent: Puppet::Provide
       return matches[:ip_dns], matches[:port], matches[:host_header]
     end
   end
+
+  def ssl?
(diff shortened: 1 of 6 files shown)


RandomNoun7:
This PR is WIP at the moment because it probably needs some additional acceptance tests.

ThoughtCrhyme (dismissed):

ThoughtCrhyme (approved):

PR #180, opened 2018-08-17
(MODULES-6870) Fix Password Escaping

RandomNoun7:
This change ensures that passwords are propertly escaped before being
used in PowerShell strings.


ThoughtCrhyme (changes requested):
The reason this PR is failing CI is because different puppet agent versions handle escape characters differently.

Puppet agent 4.10.12 (as is used in Smoke Acceptance tests) will parse a password in a manifest:
password  => Sensitive('#@\\\'454sdf')
as '#@'454sdf'

But puppet agent version of at least 5.5.2 (as is used in Full Acceptance tests) reports the same password as '#@`\`'454sdf'.

michaeltlombardi:
@ThoughtCrhyme, I think one of those reported strings is wrong? They look identical to me.

ThoughtCrhyme:
@michaeltlombardi thanks to your suggestion it's fixed now

ThoughtCrhyme (approved):

PR #184, opened 2018-08-24
(MODULES-6947) Add additional protocols

RandomNoun7:
Add support for net.tcp, net.msmq and msmq.formatname protocols. Module
currently supports only http, https and net.pipe bindings. No acceptance tests
were added. After discussion it was decided that since all possible bindings are
not currently tested, and since no changes were made to providers, only types,
that there was no real value add to adding acceptance tests for these new
protocols.


RandomNoun7:
@Pedro-Cunha This new PR has your changes, plus a couple small edits.

michaeltlombardi (approved):

PR #189, opened 2018-09-05
(MODULES-5493) Manage Two Same Name Appications

RandomNoun7:
This commit fixes a bug in the way this module uses the composite name key
feature. Prior to this change the `sitename` property was specified as
a part of the composite key. Unfortunately, properties are not valid
components of a composite key and the directive was ignored. This meant
that in truth there was only one attribute considered for the naming key,
and thus for uniqueness, which was the `applicationname` parameter.

This change removes the composite namevar functionality. The provider
has been modified such that if a user adds the site name to the applicationname
parameter, it can tell what part of the applicationname actually refers to the
sitename and handle the title accordingly.

This allows users to define applicationname's that include the sitename so
that uniqueness can be maintained for catalog compilation, but the provider
knows the name of the application to be created and isn't confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,43 @@ iis_virtual_directory { 'vdir':
 
 Allows creation of a new IIS Application and configuration of application parameters.
 
-The iis_application type uses a composite namevar for applicationname and sitename to uniquely identify a declaration. To use this successfully, put both the sitename and the applicationname in the title. Puppet will build the catalog using the composite of the two values, while still using the correct value for the applicationname when creating the IIS application.  It requires a `\` in between the `sitename` and `applicationname`, for example, `iis_application { '#{@site_name}\\#{@app_name}'`. 
+The iis_application type creates IIS Applications from directories and virtual directories, within an IIS Website. To use this type you must specify the name of a website, and then the path within the website where you would like to host an application. To do this you can use either the title only or you can use the application name and the sitename parameters separately or in combination with the title. You may also omit the sitename if you are converting an virtual directory to an application, as that parameter will already tell the provider the name of the site.
+
+```puppet
+iis_application {"$site_name\\$app_name":
+  ensure => present,
+}
+
+iis_application { $app_name:
+  ensure   => present,
+  sitename => $site_name,
+}
+
+iis_application {'myAwesomeApp':
+  ensure          => present,
+  applicationname => $app_name, # <-- Does not need to match the title
+  sitename        => $site_name
+}
+
+iis_application {'importantApplication':
+  ensure => present,
+  virtual_directory => 'IIS:\\Sites\\important_website\\importantApplication'
+}
+```
+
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -172,7 +172,43 @@ iis_virtual_directory { 'vdir':
...
+iis_application {'site1/api':
+  ensure => present,
+}
+
+iis_application {'site2/api':
+  ensure => present,
+}
+```
+
+*note: The direction of the slashes used in the title does not matter. Both directions and any combination of them are supported.
```
glennsarti: nit: `direction of the slashes` why not just be explicit.

`Both backslashes and forward slashes can be used`

README.md
```diff
@@ -172,7 +172,43 @@ iis_virtual_directory { 'vdir':
...
+
+iis_application {'myAwesomeApp':
+  ensure          => present,
+  applicationname => $app_name, # <-- Does not need to match the title
+  sitename        => $site_name
+}
+
+iis_application {'importantApplication':
+  ensure => present,
+  virtual_directory => 'IIS:\\Sites\\important_website\\importantApplication'
```
glennsarti, line 195: Given that 'IIS:\Sites\' is always required, can we remove it for the user and just automatically prefix it?

glennsarti: or alternatively just munge it on if it doesn't start with that?

glennsarti (commented):
This definitely looks like the right direction to take.

README.md
RandomNoun7: I like the munge idea. I didn't want to force paths using the IIS provider because the docs say ConvertTo-WebApplication will support 'computer name/webroot/apphost' paths, but in testing I think the docs are just wrong and paths of that format are just broken for that cmdlet. So with that in mind I think munging is fine because if it's broken for that path format then the path is just not valid.

RandomNoun7: The munge is complete.

spec/unit/puppet/type/iis_application_spec.rb
```diff
@@ -47,6 +34,15 @@
     end
     it { expect(subject[:virtual_directory]).to eq 'IIS:\Sites\foo\bar' }
   end
+  context 'specifying virtua_directory with no provider path' do
```
ThoughtCrhyme: *virtual_directory

ThoughtCrhyme:
Passing adhoc here: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/iis/job/forge-windows_puppetlabs-iis_init-manual-parameters_adhoc/85/

jpogran (approved):
yougotthepower

andrewtwemlow:
Any estimate when this will be released?

RandomNoun7:
@twem We have a ticket created for release and I expect it hit the forge in the next couple weeks.

PR #209, opened 2019-03-22
(MODULES-6080) Ensure virtualdir removal idempotent

michaeltlombardi:
This commit adds the `Recurse` parameter to calls to the
`Remove-WebVirtualDirectory` function, ensuring removals
of folders which have subfolders and child items are
actually removed.

Prior to this commit those calls would prompt a confirmation
which the run cannot give, causing failures and flapping.

### DIFF
--- a/spec/support/utilities/iis_virtual_directory.rb
+++ b/spec/support/utilities/iis_virtual_directory.rb
@@ -9,7 +9,7 @@ def create_vdir(vdir_name, site = 'foo', path = 'C:\inetpub\wwwroot')
 end
 
 def remove_vdir(vdir_name, site = 'foo', path = 'C:\inetpub\wwwroot')
-  command = format_powershell_iis_command("Remove-WebVirtualDirectory -Name #{vdir_name} -Site #{@site} -PhysicalPath #{@path}")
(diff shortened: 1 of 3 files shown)


RandomNoun7 (dismissed):

spec/support/utilities/iis_virtual_directory.rb
```diff
@@ -9,7 +9,7 @@ def create_vdir(vdir_name, site = 'foo', path = 'C:\inetpub\wwwroot')
 end
 
 def remove_vdir(vdir_name, site = 'foo', path = 'C:\inetpub\wwwroot')
-  command = format_powershell_iis_command("Remove-WebVirtualDirectory -Name #{vdir_name} -Site #{@site} -PhysicalPath #{@path}")
+  command = format_powershell_iis_command("Remove-Item -Path 'IIS:\\Sites\\#{@resource[:sitename]}\\#{@resource[:name]}' -Recurse -ErrorAction Stop")
```
RandomNoun7: Acceptance tests failing because `@resource` is not a thing in this file. Need to translate to use the `vdir_name` @site` and `@path` objects instead.

RandomNoun7 (approved):

PR #212, opened 2019-03-27
Remove redundant entry

michaeltlombardi:


michaeltlombardi:
Tried to do via UI and it derped, ignore.

PR #235, opened 2019-08-06
(MODULES-9346) - Unit tests & Acceptance tests Validation

lionce:
iis_sites
iis_applicatio*
iis_virtual

All these tests are failing because nodes are not configured in nodeset
(spec_helper_acceptance)
`install_module_from_forge_on(windows_hosts, 'puppetlabs/dism', '>= 1.2.0')
      pp = "dism { ['IIS-WebServerRole','IIS-WebServer', 'IIS-WebServerManagementTools']: ensure => present }"
      apply_manifest_on(windows_hosts, pp)`
is executed after first test is ran -> test is failing because the test environment is not setup


spec/acceptance/iis_application_pool_spec.rb
```diff
@@ -3,269 +3,235 @@
...
+        puppet_resource_should_show('load_user_profile', :false, resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('logon_type', 'LogonBatch', resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('manual_group_membership', :false, resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('max_processes', '1', resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('pinging_enabled', :true, resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('ping_interval', '00:00:30', resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('ping_response_time', '00:01:30', resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('set_profile_environment', :true, resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('shutdown_time_limit', '00:01:30', resource('iis_application_pool', pool_name))
+        puppet_resource_should_show('startup_time_limit', '00:01:30', resource('iis_application_pool', pool_name))
```
RandomNoun7: I see that you factored the global `@resource` variable out of this, and that's good, but calling the `resource` function for each of these test cases is going to drastically increase the amount of time it takes to run this test suite. Each one of those calls to `resource` has to run the same command over and over again against the test box. If instead you could do something like use `result` as a local variable and pass it instead of use it as a global, or even something like the pseudo code below, which may not be exactly syntactically correct but should give you the idea.
```
resource_data = resource('iis_application_pool', pool_name)

[
  'managed_pipeline_mode',            'Integrated', 
  'state',                            'Stopped', 
  'auto_start',                       :true,
  'enable32_bit_app_on_win64',        :false,
  'enable_configuration_override',    :true,
  'pass_anonymous_token',             :true,
  'start_mode',                       'OnDemand', 
  'queue_length',                     '1000', 
  'cpu_action',                       'NoAction', 
  'cpu_limit',                        '0', 
  'cpu_reset_interval',               '00:05:00', 
  'cpu_smp_affinitized',              :false,
  'cpu_smp_processor_affinity_mask',  '4294967295', 
  'cpu_smp_processor_affinity_mask2', '4294967295', 
  'identity_type',                    'ApplicationPoolIdentity', 
  'idle_timeout',                     '00:20:00', 
  'load_user_profile',                :false,
  'logon_type',                       'LogonBatch', 
  'manual_group_membership',          :false,
  'max_processes',                    '1', 
  'pinging_enabled',                  :true,
  'ping_interval',                    '00:00:30', 
  'ping_response_time',               '00:01:30', 
  'set_profile_environment',          :true,
  'shutdown_time_limit',              '00:01:30', 
  'startup_time_limit',               '00:01:30'
].each_slice(2) do | key, value |
  puppet_resource_should_show(key, value, resource_data)
end
```

RandomNoun7:
@lionce The problem you are running into is subtle and annoying, but there's a reason it's happening. The problem is the way that rspec does pre parsing of test files. Whenever rspec is asked to run a test file, the first thing it does read and execute the entire test file. Most of the time executing the file means reading the context blocks in a file, reading the before and after blocks in the file, and in the course of the execution generating a dependency graph for what needs to be executed in the proper order. This is why for example you can have an `after(:all)` block directly after a `before(:all)` block, and rspec will still execute the `after` in the correct order, after all of the tests are run.

The side effect though, that you are running into now, is that if you have arbitrary code that is not contained within some kind of construct that tells rspec *when* that code should run, then it will run immediately as the file is parsed and executed, and from your point of view, out of order.

Constructs that tell rspec when things should execute include `before` and `after` blocks which are self explanatory, `let`'s which tell rspec to execute a piece of code only when referred to, and others, down to `it` blocks, which are the 'examples' to run themselves. Anything that is not contained within a construct like that will be executed in an unexpected order.

If we take a look at your changes to the spec/acceptance/iis_virtual_directory_spec.rb, we'll see that a number of items have been moved out of `before` blocks, and are on their own within context blocks. Code that is on it's own in the file like that will simply be executed as they are encountered in the file, not in the proper order you would want them to be executing, which is after any specified `before` blocks, such as the `before :suite` block in the `spec_helper_acceptance`. This is why your `puts` statements on lines 23 and 41 are running before they should.

It might then be confusing why the manifest itself is running before it should, when the manifest execution is hidden inside the `it_behaves_like` function, which even in the previous version was not inside any of those constructs. The answer is that the definition of `an idempotent resource` has been changed to remove the `it` blocks within. Those `it` blocks were what ensured that the manifest was not attempted to apply until the tests themselves were running.

Notice that the code that is tasked with installing the DISM module from the forge and installing the proper Windows Server Roles is within a `before :suite` block. Since the code on line 42 is now unprotected by `it` examples, rspec is parsing `iis_virtual_directory_spec.rb`, following the code into the definition of `an idempotent resource`, and it simply executes what it finds, instead of adding the examples to the dependency graph of the tests that should be run in the correct order.

I see that you have gone to a lot of effort to remove the global variables that used to be present in the `before` blocks in these files. I think that's a worthwhile effort, however, you have to be careful not to allow any of the code you are replacing it with to run outside the protection of the constructs provided by rspec to ensure things run in the proper order. If you have code that is not protected by `before`, `after`, `let`, `it`, or a function definition called by one of those, then most likely that code will execute in an unexpected order and break the tests.

I hope this comment was helpful, and if not, please do feel free to schedule a video chat with me very early in one of our mornings and we can pair program on this for a while.

RandomNoun7 (dismissed):
I'm going to leave this comment only on this line of code, because it's particularly indicative of the issue that I'm referring to, but keep in mind that any time in these test files where you have more than one consecutive call to the `resource()` function to gather the same data will need to be modified in some way to prevent the issue. If you can find a better way than the one I have given as an example below, that's great, run with it. The main idea is to prevent multiple consecutive executions of `resource()`

lionce:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/iis/job/forge-windows_puppetlabs-iis_init-manual-parameters_adhoc/lastBuild/

spec/unit/puppet_x/puppetlabs/iis/iis_version_spec.rb
```diff
@@ -5,15 +5,16 @@
 describe PuppetX::PuppetLabs::IIS::IISVersion do
```
lionce: RSpec/FilePath:
  Enabled: false

spec/unit/puppet_x/puppetlabs/iis/iis_version_spec.rb
```diff
@@ -5,15 +5,16 @@
...
   before(:each) do
     skip 'Not on Windows platform' unless Puppet::Util::Platform.windows?
-    @ps = described_class
   end
 
   describe 'when iis is installed' do
+    let(:ps) { described_class }
+
     it 'detects a iis version' do
       expect_any_instance_of(Win32::Registry).to receive(:open)
```
lionce, line 14: I tried to replace this with

` allow(PuppetX::PuppetLabs::IIS::IISVersion::Win32::Registry).to receive(:new).and_return(registry_instance)
      allow(registry_instance).to receive(:open)`

or 
` allow(Win32::Registry).to receive(:new).and_return(registry_instance)
      allow(registry_instance).to receive(:open)`

the failing [fix](the failing fix [here](https://github.com/puppetlabs/puppetlabs-iis/pull/235/commits/1f02216112ecad18507d70abad20783588e650e4))

but tests are failing
so, i ignored rule in .rubocop.rb

RSpec/AnyInstance:
Enabled: false

lionce:
Screenshot 2019-08-26 at 20 06 45

lionce:
Screenshot 2019-08-29 at 16 37 48

lionce:
Screenshot 2019-08-29 at 16 37 48

RandomNoun7:
image

tphoney (dismissed):
I am happy with the changes, one thing would be to change the PR title :)

spec/acceptance/iis_application_spec.rb
```diff
@@ -8,423 +8,417 @@
...
     end
 
     after(:all) do
-      remove_app(@app_name)
+      remove_app(app_name)
       remove_all_sites
     end
   end
-end
+end
```
michaeltlombardi: Nit: EOF

spec/acceptance/iis_site_spec.rb
```diff
@@ -190,546 +167,517 @@
...
 
-    after(:all) do
-      remove_all_sites
+      after(:all) do
+        remove_all_sites
+      end
     end
   end
-end
+end
```
michaeltlombardi: Nit: EOF

spec/acceptance/iis_virtual_directory_spec.rb
```diff
@@ -15,226 +14,204 @@
...
 
         after(:all) do
-          remove_vdir(@virt_dir_name)
+          remove_vdir(virt_dir_name)
         end
       end
     end
   end
-end
+end
```
michaeltlombardi: Nit: EOF

spec/acceptance/iis_virtual_directory_spec_test.rb
```diff
@@ -0,0 +1,205 @@
...
+          puppet_resource_should_show('ensure', 'absent', resource('iis_virtual_directory', virt_dir_name))
+        end
+
+        after(:all) do
+          remove_vdir(virt_dir_name)
+        end
+      end
+    end
+  end
+end
```
michaeltlombardi: Nit: EOF

michaeltlombardi (dismissed):
Minor nits around EOF newlines, otherwise 👍

lionce:
Screenshot 2019-08-30 at 11 16 33

lionce:
Screenshot 2019-08-29 at 16 37 48

michaeltlombardi (approved):

PR #249, opened 2019-12-10
(FM-8195) Convert to Litmus

florindragos:


michaeltlombardi:
The second commit in this PR splits the acceptance tests for appveyor into two suites using tags; this is to ensure that the tests can actually complete in under 60m consistently, enabling them to run on each PR.

A future set of commits should refactor the test suite properly to shell out to bolt less frequently, reduce duplcate tests, and otherwise seek speed improvements so we can reconsolidate the CI setup.

DavidS (approved):
this test suite is cursed, but at least it's passing. 

Happy Holidays.

PR #252, opened 2019-12-24
(MODULES-10361) sort https bindings first

pillarsdotnet:
Fixes: [MODULES-10361](https://tickets.puppetlabs.com/browse/MODULES-10361)

When creating an `iis_site` resource, if non-ssl bindings come before ssl bindings, resource creation fails with an error similar to the following:

> `change from 'absent' to 'started' failed: Could not set 'started' on ensure: Error starting website: System.Management.Automation.CmdletInvocationException: Cannot find path 'IIS:\Sites\MyWebsite' because it does not exist.`

This patch sorts `https` bindings before non-`https` bindings.

### DIFF
--- a/lib/puppet/type/iis_site.rb
+++ b/lib/puppet/type/iis_site.rb
@@ -2,6 +2,7 @@ require 'puppet/parameter/boolean'
 require_relative '../../puppet_x/puppetlabs/iis/property/name'
 require_relative '../../puppet_x/puppetlabs/iis/property/hash'
 require_relative '../../puppet_x/puppetlabs/iis/property/path'
+require_relative '../../puppet_x/puppetlabs/iis/bindings'
 
 Puppet::Type.newtype(:iis_site) do
   @doc = "Allows creation of a new IIS Web Site and configuration of site
@@ -170,6 +171,19 @@ Puppet::Type.newtype(:iis_site) do
         value['certificatehash'] = value['certificatehash'].upcase
       end
       value
+    end
+
+    def should
+      PuppetX::PuppetLabs::IIS::Bindings.sort_bindings(super)
+    end
+
+    def should=(values)
+      super
+      @should = PuppetX::PuppetLabs::IIS::Bindings.sort_bindings(@should)
+    end
+
+    def insync?(is)
+      PuppetX::PuppetLabs::IIS::Bindings.sort_bindings(is) == should
     end
   end
 
--- a/spec/acceptance/iis_site_spec.rb
+++ b/spec/acceptance/iis_site_spec.rb
@@ -64,6 +64,10 @@ describe 'iis_site', :suite_b do
                   'protocol'             => 'http',
                 },
                 {
+                  'bindinginformation'   => '*:8088:www.puppet.local',
+                  'protocol'             => 'http',
+                },
+                {
                   'bindinginformation'   => '*:443:www.puppet.local',
                   'certificatehash'      => '#{certificate_hash}',
                   'certificatestorename' => 'MY',
@@ -120,7 +124,7 @@ describe 'iis_site', :suite_b do
               ensure               => 'started',
               # Change capitalization to see if it break idempotency
               logpath              => 'C:\\ineTpub\\logs\\NewLogFiles',
-              physicalpath         => 'C:\\ineTpub\\new',
+              physicalpath         => 'C:\\ineTpub\\new'
             }
           HERE
           apply_manifest(manifest, catch_changes: true)
--- a/lib/puppet_x/puppetlabs/iis/bindings.rb
+++ b/lib/puppet_x/puppetlabs/iis/bindings.rb
@@ -1,0 +1,13 @@
+# The Puppet Extensions Module
+module PuppetX::PuppetLabs::IIS
+  # Bindings class
+  class Bindings
+    def self.sort_bindings(b)
+      if b.nil?
+        []
+      else
+        b.sort_by { |a| ((a['protocol'] == 'https') ? '0' : '1') + a['bindinginformation'] }
+      end
+    end
+  end
+end


pillarsdotnet:
Errors appear unrelated to patch.  However, ad-hoc local testing shows this patch does not accomplish the desired objective anyway.  Trying to find out why...

pillarsdotnet:
I don't know why tests are failing, and would welcome any help in that regard.

spec/acceptance/iis_site_spec.rb
```diff
@@ -112,6 +116,7 @@
           ].each_slice(2) do |key, value|
             puppet_resource_should_show(key, value, resource_data)
           end
+          expect(resource_data.stdout).not_to match(%r{'protocol' => 'http'.*'protocol' => 'https'}m)
```
glennsarti: The resource sorting only appears to be invoked from insync.  However that's not called during a `puppet resource`.  That comes from the instances method https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/provider/iis_site/webadministration.rb#L145-L155

pillarsdotnet: Right.  First I add the test, which fails. The next commit is supposed to fix the test, so it passes.

lib/puppet/provider/iis_site/webadministration.rb
```diff
@@ -20,6 +20,8 @@
   def create
     cmd = []
 
+    @resource[:bindings] = sort_bindings(@resource[:bindings])
```
glennsarti: Instead of modifying the underlying resource, should you be using the munge method on the type?

lib/puppet/type/iis_site.rb
```diff
@@ -171,6 +171,10 @@ def insync?(is)
       end
       value
     end
+
+    def insync?(is)
+      is.sort_by { |a| ((a['protocol'] == 'https') ? '0' : '1') + a['bindinginformation'] } == should.sort_by { |a| ((a['protocol'] == 'https') ? '0' : '1') + a['bindinginformation'] }
```
glennsarti: Given the sorting code is used here and in the provider, would it be better to move that to into a module method in the puppet_x namespace e.g.  `puppetlabs-iis/lib/puppet_x/puppetlabs/iis/bindings.rb`

lib/puppet/provider/iis_site/webadministration.rb
glennsarti: Or perhaps this should be done in the PowerShell template instead of in ruby?

https://github.com/puppetlabs/puppetlabs-iis/blob/14ed33d7ba/lib/puppet/provider/templates/webadministration/bindingproperty.ps1.erb

and

https://github.com/puppetlabs/puppetlabs-iis/blob/14ed33d7ba/lib/puppet/provider/templates/webadministration/_getwebsites.ps1.erb#L32-L40

glennsarti: e.g.  https://github.com/puppetlabs/puppetlabs-iis/blob/14ed33d7ba/lib/puppet/provider/templates/webadministration/bindingproperty.ps1.erb#L4-L6 would become

``` ruby
Clear-WebConfiguration -Filter "$($website.ItemXPath)/bindings" -Force -ErrorAction Stop

<%-
ordered_bindings = resource[:bindings].sort_by { |a| ((a['protocol'] == 'https') ? '0' : '1') + a['bindinginformation'] }

  ordered_bindings.each do |bind| -%>
```

spec/acceptance/iis_site_spec.rb
glennsarti: Actually, given that it only needs to be sorted when APPLYing the bindings, not reading them, is this test valid?

lib/puppet/type/iis_site.rb
pillarsdotnet: I confess that I'm woefully ignorant about ruby code organization.

glennsarti: It's also a little dangerous but that's a Puppet thing.  The type here is depending on the Provider to have the `sort_by` method available.  And you've only implemented that on the WebAdministration provider. If this used the IISAdministration provider it would fail with method not found.

spec/acceptance/iis_site_spec.rb
pillarsdotnet: Maybe not.  Probably I should remove the test altogether.  It should fail creation anyway.

lib/puppet/provider/iis_site/webadministration.rb
pillarsdotnet: > Instead of modifying the underlying resource, should you be using the munge method on the type?

That was my first thought, but from the examples given, the munge method when applied to an array attribute acts on each element of the array, not upon the array as a whole.

pillarsdotnet: > e.g. https://github.com/puppetlabs/puppetlabs-iis/blob/14ed33d7ba/lib/puppet/provider/templates/webadministration/bindingproperty.ps1.erb#L4-L6 would become
> 
> ```ruby
> Clear-WebConfiguration -Filter "$($website.ItemXPath)/bindings" -Force -ErrorAction Stop
> 
> <%-
> ordered_bindings = resource[:bindings].sort_by { |a| ((a['protocol'] == 'https') ? '0' : '1') + a['bindinginformation'] }
> 
>   ordered_bindings.each do |bind| -%>
> ```

I'll try that next.

lib/puppet/type/iis_site.rb
pillarsdotnet: Trying it now...

spec/acceptance/iis_site_spec.rb
pillarsdotnet: done.

lib/puppet/provider/iis_site/webadministration.rb
pillarsdotnet: Trying it now.

pillarsdotnet:
Okay, now I'm mystified all over again.  Declaring an [http binding before an https binding](https://github.com/puppetlabs/puppetlabs-iis/pull/252/commits/92f8e7da0d17ab9eef57b493ca6b74c6c2cd1ed7#diff-6c70c24b4b7bc91e0ccf10c006be0f89R66-R72) fails in local testing, but succeeds in [appveyor](https://ci.appveyor.com/project/puppetlabs/puppetlabs-iis/builds/29830494).

pillarsdotnet:
Well, at least it passes test.  Trying ad-hoc local testing to see it this resolves my problem...

pillarsdotnet:
Sorting bindings on resource create/update doesn't solve my problem.  Trying again within instances and prefetch methods.

pillarsdotnet:
That fails, too.  I don't understand the failure, though.  Help, anyone?

glennsarti:
@pillarsdotnet What version of IIS and OS are you testing on locally vs Appveyor.  And don't forget you can temporarily RDP into an appveyor worker

https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

pillarsdotnet:
> https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

I've noted my versions in the [puppetlabs ticket](https://tickets.puppetlabs.com/browse/MODULES-10361), but thanks for the appveyor tip!

pillarsdotnet:
The latest commit resolves the problem in local testing.  I'm not sure how to replicate that in Appveyor.

DavidS (approved):
Thank you for your contribution!

PR #253, opened 2019-12-24
(MODULES-10362) handle authenticationinfo consistently

pillarsdotnet:
Fixes: [MODULES-10362](https://tickets.puppetlabs.com/browse/MODULES-10362)

The handling of the `authenticationinfo` field is inconsistent.  Within the [`iis_site`](https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/type/iis_site.rb#L349-L362) resource, keys are checked for validity.

Within the [`iis_application`](https://github.com/puppetlabs/puppetlabs-iis/blob/master/lib/puppet/type/iis_application.rb#L81-L88) resource, the `insync?` function is redefined to ignore element ordering.

Both functionalities should be enforced in both places.


DavidS (approved):
Looks good to me - will wait with the merge until the rebase is confirmed working by CI

PR #266, opened 2020-02-20
pdksync - Add weekly scheduled workflows

carabasdaniel:
Add weekly scheduled workflows
pdk version: `1.16.0`


michaeltlombardi (approved):

michaeltlombardi:
Merging this with failing but unrelated CI issues which will be addressed as priority. This change does not affect the codebase or tests at all.

PR #301, opened 2020-11-30
MODULES-10884 Allow a UNC path as the physical path of a web site

palintir:
… as the physical path of a web site


carabasdaniel (approved):
LGTM 👍

PR #318, opened 2021-04-26
Using pdk 2.1.0.0 tagged version for release workflow

pmcmaw:
This change will be overridden on the next pdksync, but that is fine as I have a PR up include this change. Just speeding things up to get this module released today.

Change is here: https://github.com/puppetlabs/pdk-templates/pull/430

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,10 +38,10 @@ jobs:
           ref: ${{ github.ref }}
           clean: true
       - name: "PDK Build"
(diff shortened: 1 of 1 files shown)


carabasdaniel:
Using PDK image 2.1.0.0 seems to resolve permissions issue that have shown up when using the PDK nightly image in the previous release failure: https://github.com/puppetlabs/puppetlabs-iis/runs/2438292007?check_suite_focus=true#step:4:6

carabasdaniel (approved):

PR #325, opened 2021-05-25
Correct error handling in require ruby-pwsh

benningm:
The error handling around the require for 'ruby-pwsh'
tries to rescue with nil, but does not pass an argument
to rescue which by default will rescue only StandardError.

In this case it should rescue the LoadError which is not
a StandardError:

  LoadError.is_a? StandardError => false


michaeltlombardi (dismissed):
Awesome! Thank you for this patch!

michaeltlombardi:
@benningm - if you could sign the CLA, we'd love to merge this PR!

daianamezdrea:
@benningm, as we promised, I'll push the merge button ! Thank you for your contribution ! Cheers!

PR #336, opened 2022-03-11
(MODULES-11188) Fix physicalPath on apps and sites

chelnak:
# Context

Prior to this PR it looks like we'd chosen the incorrect property to pull back from the iis config. It turns out that `PhysicalPath` != `physicalPath` and the former was behaving differently between the app and site resources.

# What has changed?

This PR standardises on physicalPath which returns the correct value for both app and site resources. For clarity, physicalPath returns a string which we can directly manipulate if required.


david22swan (approved):
LGTM

Failure's look to be transient

PR #350, opened 2022-09-15
(MAINT) Drop support for Windows Server 2008 R2.

jordanbreen28:
Prior to this commit the metadata.json file for this module listed OSs that are not supported by puppet agent as supported. This commit aims to refactor the metadata.json file to only list OSs that are supported by puppet agent.

In this commit:
- Support was removed for Windows Server 2008 R2.

The list of supported OSs can be found here:
https://puppet.com/docs/pe/2021.7/supported_operating_systems.html

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -2,7 +2,7 @@
   "name": "puppetlabs-iis",
   "version": "8.1.0",
   "author": "puppetlabs",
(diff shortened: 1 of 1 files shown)


GSPatton (approved):
All looks good

PR #375, opened 2023-07-17
(CONT-1212) - Fixing CI pipeline for Windows-2019

Ramesh7:
## Summary

Fixing CI Pipeline failures for Win-2019.

## Additional Context
- N/A

## Related Issues (if any)
Win-2019 need reboot after enabling `Web-Server` feature.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified. (For example `puppet apply`)

### DIFF
--- a/spec/spec_helper_acceptance_local.rb
+++ b/spec/spec_helper_acceptance_local.rb
@@ -11,11 +11,50 @@ class LitmusHelper
   include PuppetLitmus
 end
 
+# This method allows a block to be passed in and if an exception is raised
+# that matches the 'error_matcher' matcher, the block will wait a set number
+# of seconds before retrying.
+# Params:
+# - max_retry_count - Max number of retries
+# - retry_wait_interval_secs - Number of seconds to wait before retry
+# - error_matcher - Matcher which the exception raised must match to allow retry
+# Example Usage:
+# retry_on_error_matching(3, 5, /OpenGPG Error/) do
+#   apply_manifest(pp, :catch_failures => true)
+# end
+def retry_on_error_matching(max_retry_count = 3, retry_wait_interval_secs = 5, error_matcher = nil)
+  try = 0
+  begin
+    try += 1
+    yield
+  rescue StandardError => e
+    raise unless try < max_retry_count && (error_matcher.nil? || e.message =~ error_matcher)
+
+    sleep retry_wait_interval_secs
+    retry
+  end
+end
+
 RSpec.configure do |c|
   # Configure all nodes in nodeset
   c.before :suite do
     # Install IIS and required features on the target host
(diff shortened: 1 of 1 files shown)


spec/spec_helper_acceptance_local.rb
```diff
@@ -11,11 +11,51 @@ class LitmusHelper
...
-    LitmusHelper.instance.run_shell('Install-WindowsFeature -name Web-Server -IncludeManagementTools') unless ENV['TARGET_HOST'].nil? || ENV['TARGET_HOST'] == 'localhost'
+    unless ENV['TARGET_HOST'].nil? || ENV['TARGET_HOST'] == 'localhost'
+      LitmusHelper.instance.run_shell('Install-WindowsFeature -name Web-Server -IncludeManagementTools')
+      LitmusHelper.instance.run_shell('Install-WindowsFeature -Name Web-HTTP-Errors')
+      result = LitmusHelper.instance.run_shell('(Enable-WindowsOptionalFeature -Online -FeatureName IIS-WebServerRole -NoRestart).RestartNeeded')
+
+      if result.stdout.split("\r\n").last == 'True'
+        puts 'VM need reboot, hence doing force reboot'
+        LitmusHelper.instance.run_shell('Restart-Computer -Force')
+        sleep(300)
```
jordanbreen28: Looks like redundant use of sleep. `retry_on_error_matching` is used below to wait for the VM to come back online, so no need for both.
 Can probably remove this line `sleep(300)`
Just curious on your thoughts here

metadata.json
```diff
@@ -17,8 +17,6 @@
     {
       "operatingsystem": "Windows",
       "operatingsystemrelease": [
-        "2012",
-        "2012 R2",
```
jordanbreen28: I don’t think we can drop support whilst agent still supports these platforms.
If it is the case I think we should at least have this in a separate PR so it has a dedicated change log entry/release note.

Ramesh7: I have removed it as we don't have image available, here is from nightly build :
```
Warning: Cannot find image for Windows-2012
Warning: Cannot find image for Windows-2012 R2
```

spec/spec_helper_acceptance_local.rb
Ramesh7: The idea was I don't wanted to immediately try to run command as system just shutdown and its gonna take time to get it back, so having upfront 5 min sleep make sense and then we can have small retries like 5 sec with 60 or etc.

metadata.json
jordanbreen28: Yeah I think like you noted we need to align our support, or atleast find suitable images to run the tests on

jordanbreen28 (approved):

PR #380, opened 2023-09-11
[CAT-1382] : Fixing CI failure due to rubocop

rajat-puppet:
## Summary
Fixing CI failure due to rubocop

## Additional Context
- [x] New version released in rubocop-rspec ([v2.24.0](https://github.com/rubocop/rubocop-rspec/releases/tag/v2.24.0)) caused regression in linting.
```
Split RSpec/FilePath into RSpec/SpecFilePathSuffix and RSpec/SpecFilePathFormat. RSpec/FilePath cop is enabled by default, the two new cops are pending and need to be enabled explicitly. (@ydah)
Add RSpec/MetadataStyle and RSpec/EmptyMetadata cops.
```

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -98,13 +98,6 @@ RSpec/ExampleLength:
   Max: 37
 
 # Offense count: 1
-# Configuration parameters: Include, CustomTransform, IgnoreMethods, SpecSuffixOnly.
-# Include: **/*_spec*rb*, **/spec/**/*
-RSpec/FilePath:
-  Exclude:
-    - 'spec/acceptance/iis_virtual_directory_spec_test.rb'
-
-# Offense count: 1
 # Configuration parameters: AssignmentOnly.
 RSpec/InstanceVariable:
   Exclude:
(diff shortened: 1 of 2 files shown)


Ramesh7 (dismissed):

.rubocop_todo.yml
```diff
@@ -140,6 +133,7 @@ RSpec/NoExpectationExample:
     - 'spec/unit/puppet/provider/iis_virtual_directory/webadministration_spec.rb'
     - 'spec/unit/puppet/type/iis_site_spec.rb'
     - 'spec/unit/puppet/type/iis_virtual_directory_spec.rb'
+    - 'spec/acceptance/iis_virtual_directory_test_spec.rb'
```
Ramesh7, line 136: we should remove `'spec/acceptance/iis_virtual_directory_spec_test.rb` from above list as well?

rajat-puppet: This will remain here because for each expect block we have a method written instead of explicitly calling expect inside each it block of specs.

Ramesh7 (approved):

PR #406, opened 2025-06-03
fix(MODULES-5225): include app_name in cache key for virt_dir_path

imaqsood:
## Summary
Ensure app_name is part of the cache key to prevent incorrect cache hits when the same sitename and name are used with different app_name values.

Also refactored the method for clarity and safer string handling.

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)


amitkarsale (approved):
LGTM

PR #407, opened 2025-06-03
Fix: Unordered IIS Site Binding Comparison to Prevent Unnecessary Resource Updates

DarkArc-Github:
# Fix IIS Site Binding Comparison to Prevent Unnecessary Resource Updates

## Summary

This pull request modifies the `insync?` method in the `bindings` property of the `iis_site` resource type to use a set-based comparison instead of an order-dependent array comparison. This change prevents Puppet from detecting false differences and unnecessarily reapplying bindings when the only difference is the order in which IIS returns the bindings.

The specific change:
- Modified the `insync?` method in `Puppet/iis/lib/puppet/type/iis_site.rb` to convert binding arrays to sets before comparison, making the comparison order-independent

## Additional Context

- [x] Root cause and the steps to reproduce:
  - Root cause: The current implementation compares bindings as ordered arrays, but the order of bindings returned by IIS can vary between runs, especially for bindings with the same binding information but different protocols (such as `net.msmq` and `msmq.formatname`).
  - To reproduce: Create an IIS site with multiple binding types, including at least one pair of non-HTTP bindings with the same binding information (e.g., `net.msmq` and `msmq.formatname` both using the same hostname). Run Puppet multiple times and observe that it detects changes and reapplies the bindings on every run.

- [x] Thought process behind the implementation:
  - I considered several approaches to fix this issue:
    1. Modifying the `sort_bindings` method to ensure a stable sort
    2. Creating a custom provider that overrides the binding comparison
    3. Using a set-based comparison in the `insync?` method
  - I chose the set-based approach because:
    - It's the simplest and most direct solution
    - It aligns with how IIS actually works (binding order doesn't affect functionality)
    - It's the least invasive change with the lowest risk of unintended side effects
    - It's more robust against variations in how IIS or PowerShell returns bindings

## Related Issues

This PR addresses an issue where IIS sites with multiple binding types would be unnecessarily updated on every Puppet run, causing potential service disruptions and increased Puppet run times.

## Checklist

- [x] 🟢 Spec tests: Tested locally [unit test log](https://github.com/user-attachments/files/20629196/spec_test.log)
- [x] 🟢 Acceptance tests: NA as it already tests for idempotency.
- [x] 🟢 Manually verified: Tested on a Windows Server with IIS and confirmed that after this change, Puppet correctly recognizes when bindings are in sync, even if their order varies


DarkArc-Github:
@span786 Can you please review this?

do you want me to add changelog and metadata updates as well to this PR ?

span786:
@DarkArc-Github 

Could you please add the tests?
Also, I noticed some `rubocop` violations—please address those as well.

lib/puppet/type/iis_site.rb
```diff
@@ -166,7 +166,12 @@ def should=(values)
     end
 
     def insync?(is)
-      PuppetX::PuppetLabs::IIS::Bindings.sort_bindings(is) == should
+      #Convert bindings to set of hashes for unordered comparison
+      is_set = is.to_set
+      should_set = should.to_set
+
+      #Compare the sets
+      is_set = should_set
```
span786: Please correct this line of code and add a proper comparison.

DarkArc-Github:
> @DarkArc-Github
> 
> Could you please add the tests? Also, I noticed some `rubocop` violations—please address those as well.

Thanks for the reply, I don't see the spec jobs running when I push, so I can't find those issues when I make changes, I saw that job only when you commented so can you point me in the right direction please how can I get those jobs running spec tests for my push ?

lib/puppet/type/iis_site.rb
```diff
@@ -166,7 +166,7 @@ def should=(values)
     end
 
     def insync?(is)
-      PuppetX::PuppetLabs::IIS::Bindings.sort_bindings(is) == should
+      is.to_set = should.to_set
```
span786: This should be `is.to_set == should.to_set`

span786:
> > @DarkArc-Github
> > Could you please add the tests? Also, I noticed some `rubocop` violations—please address those as well.
> 
> Thanks for the reply, I don't see the spec jobs running when I push, so I can't find those issues when I make changes, I saw that job only when you commented so can you point me in the right direction please how can I get those jobs running spec tests for my push ?

You can check the test steps, and follow the same while running tests in your local

DarkArc-Github:
@span786 Added unit test and removed the unused and unreferenced bindings class and tested locally (attached the spec run log here for your reference).
Added details to reference markdown as well for easy lookup.

- Unit Spec test passes locally. [spec_test.log](https://github.com/user-attachments/files/20629196/spec_test.log)
- Tested locally by applying on test node.

span786:
@DarkArc-Github  Please fix the `rubocop` errors.

DarkArc-Github:
@span786 Thank for the help, all spec tests are clear, its failing on acceptance test, looks like failed to provision test VM.

DarkArc-Github:
@span786 Can you please help me further on this please?

span786:
> @span786 Can you please help me further on this please?

There’s currently an issue with the VM provisioner that's causing the acceptance tests to fail. I’ll need to investigate and resolve that first before proceeding with this PR.

DarkArc-Github:
Hi @span786 ,

Just checking in if there's anything I can do to help expedite this?

span786:
The CI failures are unrelated, so I'm approving this PR.

span786 (approved):

DarkArc-Github:
@span786 Thank you!
Just for my understanding and to be to be more helpful in future, when will this be available in a new version and how is the new version provisioned?

I assume it is automated but broken because of the broken VM pipeline.

How can I be more actively involved in maintaining this or other modules that can use some help?

span786:
@DarkArc-Github 

This will be included in the next release. While provisioning of the new version is automated, it requires approval before the release is finalized. The release pipeline itself is not dependent on VM provisioning.

This is the standard process for supporting and maintaining modules: you raise a PR with the necessary changes and tests. Once the CI runs successfully and all checks pass, we merge the PR and proceed with releasing the new version.

PR #418, opened 2026-06-22
(MODULES-11595) Mark iis_application_pool password as sensitive to stop report leak

imaqsood:
## Summary

Follow-up to #412 (released in v10.1.6). That fix masked the password in the
human-readable event **message** via `should_to_s`/`is_to_s`, but the customer
reported the plaintext password **still appears in the PE console report**.

## Root cause

A Puppet change event carries structured `desired_value` / `previous_value`
fields in addition to the message. The PE console report renders those fields,
and Puppet only redacts them when the property is flagged **sensitive**
(`Puppet::Transaction::ResourceHarness#create_change_event`):

```ruby
if property.sensitive
  options[:previous_value] = current_value.nil? ? nil : '[redacted]'
  options[:desired_value]  = should.nil? ? nil : '[redacted]'
else
  options[:previous_value] = current_value   # raw password
  options[:desired_value]  = should          # raw password
```

The `password` property was never marked sensitive, so the structured values
kept leaking even though the message was masked.

## Changes

- **type**: declare `sensitive true` on the `password` property. This redacts
  the structured event values **and** the message wherever a change event is
  recorded, with no `Sensitive()` wrapper required from the user. The existing
  `should_to_s`/`is_to_s` overrides are kept as defense-in-depth.
- **provider**: redact sensitive property values from the `Puppet.debug`
  "Changing ... to ..." line, and add a `redact_sensitive` helper that scrubs
  sensitive values out of PowerShell error messages before `Puppet.err`, since
  a failing command can echo the password back.
- **spec**: add coverage for the `sensitive` flag and the log redaction.

## Testing

```
bundle exec rspec spec/unit/puppet/provider/iis_application_pool/webadministration_spec.rb \
                  spec/unit/puppet/type/iis_application_pool_spec.rb
# 86 examples, 0 failures
bundle exec rubocop <changed files>   # no offenses
```

Jira: https://perforce.atlassian.net/browse/MODULES-11595

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/puppet/provider/iis_application_pool/webadministration.rb
+++ b/lib/puppet/provider/iis_application_pool/webadministration.rb
@@ -40,7 +40,8 @@ Puppet::Type.type(:iis_application_pool).provide(:webadministration, parent: Pup
 
     @resource.properties.select { |rp| rp.name != :ensure && rp.name != :state }.each do |property|
       property_name = iis_properties[property.name.to_s]
-      Puppet.debug "Changing #{property_name} to #{property.value}"
+      log_value = property.sensitive ? '[redacted]' : property.value
+      Puppet.debug "Changing #{property_name} to #{log_value}"
       if property.value.is_a?(Array)
         cmd << "Clear-ItemProperty -Path 'IIS:\\AppPools\\#{@resource[:name]}' -Name '#{property_name}'"
         property.value.each do |item|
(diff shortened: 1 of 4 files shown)


imaqsood:
Thanks for the review — all three points were valid and are addressed in 65cc5e4:

1. **Debug-level leak in the shared `run`** (the important one): `IIS_PowerShell.run` logged `COMMAND: <full command>` / `STDOUT` / `ERRMSG` at debug *before* the provider could redact anything, so the password still leaked with debug logging on. `run` now accepts an optional `sensitive_values:` list and scrubs those values from every debug line, while still executing the command verbatim. `iis_application_pool#update` passes the values of all `sensitive` properties.

2. **Duplicate / incorrect `Puppet.err`**: it previously logged twice on a real failure and logged a spurious error on success when `errormessage` was an empty string (`'' != nil`). It now logs once, only on an actual failure (`unless exitcode.zero? && errormessage blank`).

3. **Spec stubbed `run` and missed the real leak**: added a `.run` spec that stubs `ps_manager` instead of `run`, so the actual `COMMAND:` redaction path is exercised, plus an assertion that the secret is still passed verbatim to `execute`, and a case for the successful-empty-`errormessage` path.

89 specs passing, rubocop clean.

span786 (approved):

